PR #10, opened 2018-03-05
fix compilation on non-glibc linux

jsteemann:
This should fix compilation of s2geometry on Linux when there is no glibc around.
Compilation failed because macro `IS_BIG_ENDIAN` or `IS_LITTLE_ENDIAN` do not seem to get defined in some environments:
```
/work/ArangoDB/3rdParty/s2geometry/src/s2/util/endian/endian.h:111:2: error: #error "Unsupported bytesex: Either IS_BIG_ENDIAN or IS_LITTLE_ENDIAN must be defined"
 #error "Unsupported bytesex: Either IS_BIG_ENDIAN or IS_LITTLE_ENDIAN must be defined"  // NOLINT
  ^~~~~
```
This seems to be because the define `OS_LINUX` also is not defined in this environment.
So I am now checking for the define `__linux__` and not just `OS_LINUX` to detect a Linux OS.

### DIFF
--- a/src/s2/base/port.h
+++ b/src/s2/base/port.h
@@ -159,7 +159,7 @@
 #endif  // defined(__APPLE__)
 
 // __GLIBC_PREREQ
-#if defined OS_LINUX
+#if defined OS_LINUX || defined(__linux__)
 // GLIBC-related macros.
 #include <features.h>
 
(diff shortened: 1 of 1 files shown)


jmr:
[I just replaced OS_LINUX with __linux__](https://github.com/google/s2geometry/commit/2bfcfc690366cdeed2202ef069e2f0501b993443). Try that and let me know if it works for you. Thanks for the report.

jsteemann:
@jmr : yes, this works too. Thanks for the fix!

PR #14, opened 2018-03-26
fix ubuntu dependency instructions

bdon:
the library name for openssl headers/libraries is `libssl-dev` and not `openssl` (which is the executables)


jmr:
If you sign the CLA, I can merge this. It should just be a few clicks.

bdon:
OK, i just did that.

PR #17, opened 2018-04-02
Improve portability and reduce warnings

danielhlarkin:
We've been working on integrating S2 into our project for a little while, and we've found a few issues with portability (especially compiling on Windows), and with some extraneous compile warnings (e.g. with gcc-7.2). This PR contains a mish-mash of small changes to address these concerns, hopefully in a way that can benefit everyone.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,7 +54,7 @@ endif()
 
 add_definitions(-std=c++11)
 
-if(WIN32)
+if (WIN32)
     # Use unsigned characters
     add_definitions(-J)
     # Make sure cmath header defines things like M_PI
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -156,6 +156,7 @@ add_library(s2 ${LIBRARY_TYPE}
             src/s2/s2shapeutil_visit_crossing_edge_pairs.cc
             src/s2/s2text_format.cc
             src/s2/s2wedge_relations.cc
+            src/s2/strings/ostringstream.cc
             src/s2/strings/serialize.cc
             src/s2/third_party/absl/base/dynamic_annotations.cc
             src/s2/third_party/absl/base/internal/raw_logging.cc
@@ -296,6 +297,8 @@ install(FILES src/s2/base/casts.h
               src/s2/base/port.h
               src/s2/base/spinlock.h
         DESTINATION include/s2/base)
+install(FILES src/s2/strings/ostringstream.h
+        DESTINATION include/s2/strings)
 install(FILES src/s2/third_party/absl/algorithm/algorithm.h
         DESTINATION include/s2/third_party/absl/algorithm)
 install(FILES src/s2/third_party/absl/base/attributes.h
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -61,8 +61,6 @@ if(WIN32)
     add_definitions(-D_USE_MATH_DEFINES)
     # Make sure Windows doesn't define min/max macros that interfere with STL
     add_definitions(-DNOMINMAX)
-    # Add a flag to define ssize_t for btree implementation
-    add_definitions(-DNEED_ALT_SSIZE)
 else()
     add_definitions(-Wno-deprecated-declarations)
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,10 +17,6 @@ cmake_dependent_option(WITH_GLOG "Use glog for logging." ON
                        "WITH_GFLAGS" OFF)
 add_feature_info(GLOG WITH_GLOG "provides logging configurability.")
 
-option(BUILD_SHARED_LIBS "Build shared libraries instead of static." ON)
-add_feature_info(SHARED_LIBS BUILD_SHARED_LIBS
-                 "builds shared libraries instead of static.")
-
 feature_summary(WHAT ALL)
 
 if (WITH_GLOG)
@@ -73,7 +69,7 @@ include_directories(
     ${PYTHON_INCLUDE_DIRS})
 include_directories(src)
 
-add_library(s2
+add_library(s2 SHARED
             src/s2/base/stringprintf.cc
             src/s2/base/strtoint.cc
             src/s2/id_set_lexicon.cc
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,7 +55,14 @@ endif()
 add_definitions(-std=c++11)
 
 if(WIN32)
-    add_definitions(-J -D_USE_MATH_DEFINES -DNOMINMAX -DNEED_ALT_SSIZE)
+    # Use unsigned characters
+    add_definitions(-J)
+    # Make sure cmath header defines things like M_PI
+    add_definitions(-D_USE_MATH_DEFINES)
+    # Make sure Windows doesn't define min/max macros that interfere with STL
+    add_definitions(-DNOMINMAX)
+    # Add a flag to define ssize_t for btree implementation
+    add_definitions(-DNEED_ALT_SSIZE)
 else()
     add_definitions(-Wno-deprecated-declarations)
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,6 +17,10 @@ cmake_dependent_option(WITH_GLOG "Use glog for logging." ON
                        "WITH_GFLAGS" OFF)
 add_feature_info(GLOG WITH_GLOG "provides logging configurability.")
 
+option(STATIC_LIB "Build static library instead of shared." OFF)
+add_feature_info(STATIC_LIBRARY STATIC_LIB
+                 "builds static library instead of shared.")
+
 feature_summary(WHAT ALL)
 
 if (WITH_GLOG)
@@ -40,6 +44,12 @@ if (WITH_GFLAGS)
     add_definitions(-DS2_USE_GFLAGS)
 endif()
 
+if (STATIC_LIB)
+  set(LIBRARY_TYPE "STATIC")
+else()
+  set(LIBRARY_TYPE "SHARED")
+endif()
+
 find_package(OpenSSL REQUIRED)
 # pthreads isn't used directly, but this is still required for std::thread.
 find_package(Threads REQUIRED)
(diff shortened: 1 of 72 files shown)


danielhlarkin:
CLA signed.

src/s2/s2boolean_operation.h
```diff
@@ -26,6 +26,10 @@
 #include "s2/s2builder_layer.h"
 #include "s2/value_lexicon.h"
 
+#ifdef DIFFERENCE // Windows
```
jmr: What is DIFFERENCE on Windows? Maybe we should rename this one instead. @ericveach

CMakeLists.txt
```diff
@@ -17,6 +17,10 @@ cmake_dependent_option(WITH_GLOG "Use glog for logging." ON
                        "WITH_GFLAGS" OFF)
 add_feature_info(GLOG WITH_GLOG "provides logging configurability.")
 
+option(BUILD_SHARED_LIBS "Build shared libraries instead of static." ON)
```
jmr: This could be a separate PR.

jmr: Add comment that this is OFF for Windows (if that's indeed the case -- not sure if you're adding it for Windows, or just for consistency with normal CMake practice).

src/s2/s2coords.h
```diff
@@ -335,7 +335,7 @@ inline int STtoIJ(double s) {
 }
 
 inline double SiTitoST(unsigned int si) {
-  S2_DCHECK(si >= 0 && si <= kMaxSiTi);
+  S2_DCHECK(si <= kMaxSiTi);
```
jmr: Separate "Fix unsigned warnings" PR.

src/s2/s2polyline.cc
```diff
@@ -50,6 +50,13 @@ static const unsigned char kCurrentLosslessEncodingVersionNumber = 1;
 S2Polyline::S2Polyline()
   : s2debug_override_(S2Debug::ALLOW) {}
 
+S2Polyline::S2Polyline(S2Polyline&& other)
```
jmr: Does this have anything to do with Windows?

src/s2/util/math/vector.h
```diff
@@ -175,8 +175,7 @@ class BasicVector {
 
   // Euclidean norm. For integer T, correct only if Norm2 does not overflow.
   FloatType Norm() const {
-    using std::sqrt;
-    return sqrt(Norm2());
+    return std::sqrt(Norm2());
```
jmr: Are these changes to fix errors, warnings, or style?

src/s2/util/gtl/btree.h
```diff
@@ -73,6 +73,14 @@
 #include "s2/third_party/absl/strings/string_view.h"
 #include "s2/util/gtl/subtle/compressed_tuple.h"
 
+#if defined(_WIN32) || defined(_WIN64)
```
jmr: Could you change size_type to std::make_signed<size_t>::type instead?

I'd like to keep any platform-specific code to a minimum and confine to s2/base if possible.

src/s2/s2testing.cc
```diff
@@ -320,9 +325,21 @@ void S2Testing::CheckCovering(const S2Region& region,
 }
 
 double S2Testing::GetCpuTime() {
+#if defined(_WIN32) || defined(_WIN64)
+  FILETIME creation, exit, kernel, user;
+  int res = GetProcessTimes(GetCurrentProcess(), &creation,
+                            &exit, &kernel, &user);
+  if (res == 0) {  // something went wrong
+    return 0.0;    // is this right for an error case?
```
jmr: The non-windows version just S2_CHECKs.

src/s2/s2testing.cc
```diff
@@ -320,9 +325,21 @@ void S2Testing::CheckCovering(const S2Region& region,
...
+  }
+  // high and low combine to 64-bit int, time uses 100-nanosecond intervals
+  return static_cast<double>((static_cast<uint64_t>(user.dwHighDateTime) << 32)
+                            + static_cast<uint64_t>(user.dwLowDateTime)) / 1e7;
+#else
   struct rusage ru;
```
jmr: I'm going to move this to something like base/sysinfo, so we don't need ifdef _WIN32 in core s2 code.

src/s2/s2boolean_operation.h
danielhlarkin: I don't currently have access to the headers to link/paste here, but I believe it was for something intended to be used primarily inside the header itself, some kind of API versioning deal. It shouldn't be a problem to undef it here, but you could change your own enum instead if you wish.

CMakeLists.txt
danielhlarkin: It is on for all platforms by default. The goal is to allow a simple flag to enable static builds instead of requiring a source change (as they were hardcoded SHARED) before. Will separate into another PR.

src/s2/s2coords.h
danielhlarkin: Agreed.

src/s2/s2polyline.cc
danielhlarkin: No, will separate into another PR. It was to silence a warning.

src/s2/util/math/vector.h
danielhlarkin: Warnings about mismatched attributes. There are still some others elsewhere in the codebase (and also in the Abseil code), but they did not have the same simple fix.

src/s2/util/gtl/btree.h
danielhlarkin: Will check.

danielhlarkin:
@jmr I'll separate this into multiple smaller PRs as requested. Should I keep the main Windows-compatibility changes here and move everything else?

jmr:
Keep anything with ifdef _WIN32 here, but split out Windows-inspired fixes such as make_signed.

danielhlarkin:
@jmr I've separated most of the changes out into other branches and will open new PRs momentarily. This (now much smaller) PR now does just two things:
- Adds/removes compile flags/definitions for Windows builds
- Moves `S2Testing::GetCpuTime` to `base` and makes it work on Windows

README.md
```diff
@@ -16,7 +16,6 @@ S2 documentation can be found on [s2geometry.io](http://s2geometry.io).
 
 ## Requirements for End Users
 
-* A POSIX system (for getrusage).
```
jmr: Maybe say "A POSIX system (for getrusage) or Windows."

CMakeLists.txt
```diff
@@ -52,7 +52,19 @@ if (APPLE)
     set(CMAKE_MACOSX_RPATH TRUE)
 endif()
 
-add_definitions(-Wno-deprecated-declarations -std=c++11)
+add_definitions(-std=c++11)
+
+if(WIN32)
```
jmr: `if (WIN32)` for consistency.

README.md
danielhlarkin: Fixed.

jmr:
I updated sysinfo and s2testing in a way that makes things match google code. After you merge, this should be good to go.

src/s2/base/sysinfo.cc
```diff
@@ -15,15 +15,29 @@
...
+#endif
 
 namespace base {
 
 absl::Duration CPUUsage() {
+#ifdef _MSC_VER /* if Visual C++ */
+  FILETIME creation, exit, kernel, user;
+  int res = GetProcessTimes(GetCurrentProcess(), &creation,
+                            &exit, &kernel, &user);
+  S2_CHECK_EQ(res, 0);
```
jmr: The `CHECK` stayed in `s2testing.cc`, so just `if (res != 0) return absl::ZeroDuration();` here. (Or `if (GetProcessTimes(...) != 0) return ...;`.

src/s2/base/sysinfo.cc
```diff
@@ -15,15 +15,29 @@
...
+  // high and low combine to 64-bit int, time uses 100-nanosecond intervals
+  return static_cast<double>((static_cast<uint64_t>(user.dwHighDateTime) << 32)
+                            + static_cast<uint64_t>(user.dwLowDateTime)) / 1e7;
```
jmr: Stack overflow seems to say copying high/low to a `ULARGE_INTEGER` and using `QuadPart` is the preferred way.

jmr: Add comment that the units are 100 ns intervals. 1e-7 wasn't obvious to me.

src/s2/base/sysinfo.cc
```diff
@@ -15,15 +15,29 @@
...
 namespace base {
 
 absl::Duration CPUUsage() {
+#ifdef _MSC_VER /* if Visual C++ */
+  FILETIME creation, exit, kernel, user;
+  int res = GetProcessTimes(GetCurrentProcess(), &creation,
+                            &exit, &kernel, &user);
+  S2_CHECK_EQ(res, 0);
+  // high and low combine to 64-bit int, time uses 100-nanosecond intervals
+  return static_cast<double>((static_cast<uint64_t>(user.dwHighDateTime) << 32)
```
jmr: Can you test this one final time? I think it needs to be `return absl::Duration(...);` now.

CMakeLists.txt
```diff
@@ -56,7 +56,19 @@ if (APPLE)
     set(CMAKE_MACOSX_RPATH TRUE)
 endif()
 
-add_definitions(-Wno-deprecated-declarations -std=c++11)
+add_definitions(-std=c++11)
+
+if (WIN32)
+    # Use unsigned characters
+    add_definitions(-J)
```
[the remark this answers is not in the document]
danielhlarkin: Yes, CMake takes care of converting the `-flag` to `/flag` for MSVC.

jmr:
Did you try the tests on Windows?

Could you make another PR adding yourself to the CONTRIBUTORS and AUTHORS files?

Thanks again!

danielhlarkin:
Sure, I'll add myself to the lists. I haven't explicitly run the tests on Windows, but everything does compile fine. Let me see if I can enlist someone's assistance here in running the tests...

PR #18, opened 2018-04-03
Add missing move constructor to S2Polyline

danielhlarkin:


src/s2/s2polyline.h
```diff
@@ -46,6 +46,7 @@ class S2Polyline final : public S2Region {
   // Creates an empty S2Polyline that should be initialized by calling Init()
   // or Decode().
   S2Polyline();
+  S2Polyline(S2Polyline&&);
```
jmr: Do we also need to add a move `operator=`?

src/s2/s2polyline.cc
```diff
@@ -50,6 +50,13 @@ static const unsigned char kCurrentLosslessEncodingVersionNumber = 1;
 S2Polyline::S2Polyline()
   : s2debug_override_(S2Debug::ALLOW) {}
 
+S2Polyline::S2Polyline(S2Polyline&& other)
+ : s2debug_override_(other.s2debug_override_),
+   num_vertices_(other.num_vertices_),
+   vertices_(std::move(other.vertices_)) {
+  other.num_vertices_ = 0;
```
jmr: It might be safer/more obvious to do `other_.vertices_.clear()` to get back to the normal empty state.

danielhlarkin: I don't think it's necessary to add a `release()` operation here. It's explicit in the standard that moving a `unique_ptr` has the same effect as calling `release()` on the source operand.

src/s2/s2polyline.h
```diff
@@ -47,6 +47,10 @@ class S2Polyline final : public S2Region {
   // or Decode().
   S2Polyline();
 
+// S2Polyline is movable, but only privately copyable.
```
jmr: Sorry, I wasn't clear -- with normal two-space indentation.

danielhlarkin: Ah, crap. My bad, the lack of indentation was unintentional.

src/s2/s2polyline.h
```diff
@@ -47,6 +47,10 @@ class S2Polyline final : public S2Region {
   // or Decode().
   S2Polyline();
 
+  // S2Polyline is movable, but only privately copyable.
+  S2Polyline(S2Polyline&&);
+  S2Polyline& operator=(S2PolyLine&&);
```
jmr: s/Line/line/

PR #20, opened 2018-04-03
Fix some attribute mismatch warnings.

danielhlarkin:
I am using gcc-7.2.0 on Ubuntu 17.10. This fixes a large number of warnings of the following form:
```
/home/dan/work/s2geometry/src/s2/util/math/vector.h:530:16: warning: optimization attribute on ‘double fabs(double)’ follows definition but the attribute doesn’t match [-Wattributes]
     using std::fabs;
                ^~~~
In file included from /home/dan/work/s2geometry/src/s2/third_party/absl/numeric/int128.h:29:0,
                 from /home/dan/work/s2geometry/src/s2/util/bits/bits.h:49,
                 from /home/dan/work/s2geometry/src/s2/s2cell_id.h:30,
                 from /home/dan/work/s2geometry/src/s2/mutable_s2shape_index.h:33,
                 from /home/dan/work/s2geometry/src/s2/s2builder.h:31,
                 from /home/dan/work/s2geometry/src/s2/s2boolean_operation.h:24,
                 from /home/dan/work/s2geometry/src/s2/s2boolean_operation.cc:65:
/usr/include/c++/7/cmath:245:3: note: previous definition of ‘constexpr long double std::fabs(long double)’ was here
   fabs(long double __x)
   ^~~~
In file included from /home/dan/work/s2geometry/src/s2/r2.h:22:0,
                 from /home/dan/work/s2geometry/src/s2/s2cell_id.h:33,
                 from /home/dan/work/s2geometry/src/s2/mutable_s2shape_index.h:33,
                 from /home/dan/work/s2geometry/src/s2/s2builder.h:31,
                 from /home/dan/work/s2geometry/src/s2/s2boolean_operation.h:24,
                 from /home/dan/work/s2geometry/src/s2/s2boolean_operation.cc:65:
```

### DIFF
--- a/src/s2/util/math/vector.h
+++ b/src/s2/util/math/vector.h
@@ -175,8 +175,7 @@ class BasicVector {
 
   // Euclidean norm. For integer T, correct only if Norm2 does not overflow.
   FloatType Norm() const {
-    using std::sqrt;
-    return sqrt(Norm2());
+    return std::sqrt(Norm2());
   }
 
   // Normalized vector if the norm is nonzero. Not for integer types.
@@ -192,8 +191,7 @@ class BasicVector {
   // Compose a vector from the sqrt of each component.
   D Sqrt() const {
     return Generate([](const T& x) {
-      using std::sqrt;
-      return sqrt(x);
+      return std::sqrt(x);
     }, AsD());
   }
 
@@ -342,8 +340,8 @@ class Vector2
   T y() const { return c_[1]; }
 
   bool aequal(const Vector2 &vb, FloatType margin) const {
-    using std::fabs;
-    return (fabs(c_[0]-vb.c_[0]) < margin) && (fabs(c_[1]-vb.c_[1]) < margin);
+    return (std::fabs(c_[0]-vb.c_[0]) < margin) &&
+           (std::fabs(c_[1]-vb.c_[1]) < margin);
   }
 
   void Set(T x, T y) { *this = Vector2(x, y); }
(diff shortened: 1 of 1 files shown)


ericveach:
Do you know what is causing these mismatches?  I don't even see where the optimize attribute is being declared.

danielhlarkin:
@ericveach The full definition in `/usr/include/c++/7/cmath` is
```
  inline _GLIBCXX_CONSTEXPR long double
  fabs(long double __x)
  { return __builtin_fabsl(__x); }
```
I won't claim to fully understand what's going wrong with the aliasing, but it appears to be dropping the `constexpr` somehow.

jmr:
I'm trying to construct a simple test case to reproduce this and nothing I'm doing is triggering the warning. with g++ (Debian 7.3.0-5) 7.3.0

danielhlarkin:
Hm. Interesting that 7.3.0 does not exhibit the warnings. I've tested locally and can say that 6.4.0 does not either. I believe this is likely an issue with gcc which was fixed with 7.3 (sot it may also be present in 7.0 and 7.1). Do you have an easy way to install 7.2 on Debian? I think I can probably find a 7.3 PPA for Ubuntu if necessary.

jmr:
7.3.0 does give the warnings when I compile s2.  When I try something simple like
```c++
#include <cmath>

long double f(long double x) {
  using std::sqrt;
  return sqrt(x);
}
```
I don't get any warnings. I'll have to try more to reduce what's happening in s2 to a small test case.

jmr:
This isn't a problem for `Vector`'s use in s2, but I think the version in the PR breaks ADL.  That's probably why these were written with the using statements. Is there a `-wsomething` to turn off this warning?

jmr:
Fixing with cac8a3f instead.

PR #21, opened 2018-04-03
Remove platform-specific ssize_t from btree implementation.

danielhlarkin:
Since Windows doesn't have `ssize_t`, converting to the standardized `std::make_signed<size_t>::type` fixes compilation issues there.

### DIFF
--- a/src/s2/util/gtl/btree.h
+++ b/src/s2/util/gtl/btree.h
@@ -248,7 +248,7 @@ struct common_params {
 
   using allocator_type = Alloc;
   using key_type = Key;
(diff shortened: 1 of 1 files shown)


src/s2/util/gtl/btree.h
```diff
@@ -815,7 +815,7 @@ class btree {
   };
 
   struct node_stats {
-    node_stats(ssize_t l, ssize_t i)
```
jmr: Could you do `using size_type = typename Params::size_type;` and use that instead of repeating the `typename ...`?

PR #24, opened 2018-04-03
Undefine conflicting Windows macro

danielhlarkin:
For some reason, someone thought it was a good idea to define a `DIFFERENCE` macro in a Windows header. Therefore, to avoid conflict when compiling on Windows, I've renamed `OpType::DIFFERENCE` to `OpType::SET_DIFFERENCE`.

### DIFF
--- a/src/s2/base/port.h
+++ b/src/s2/base/port.h
@@ -57,6 +57,7 @@
 #include <windows.h>
 #undef ERROR
 #undef DELETE
+#undef DIFFERENCE
 #define STDIN_FILENO 0
 #define STDOUT_FILENO 1
 #define STDERR_FILENO 2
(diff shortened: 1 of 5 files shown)


jmr:
@ericveach what do you think of SET_DIFFERENCE vs #undef DIFFERENCE for Windows?

ericveach:
Sorry for not replying about this earlier.  I'd rather stick with
DIFFERENCE (of course it's actually S2BooleanOperation::OpType::DIFFERENCE).
These are not really set operations anyway (you can think of them that way,
but it's more complicated than that).

The main downside of the #undef is that it pollutes our source files.
Which transitively included header file is it coming from?  Could we #undef
it immediately after that file is included?

danielhlarkin:
@ericveach Yes, I believe we can do that. Moving the #undef right after the include in `src/s2/base/port.h` seems to handle it. Good call.

PR #28, opened 2018-04-17
Make building examples optional

graetzer:
Currently using the provided cmake file always builds the examples. We include S2 in a larger project and we thought that it would be a good idea to make that optional via a flag


graetzer:
CLA should be signed

Issue #13, opened 2018-03-25
Warnings with GCC 5.4


graetzer:
I noticed quite a lot of warnings in clang due to signed / unsigned integer comparison as well as structs that were  forward declared as classes.

Would it be ok to submit a pull request which silences some of these ?

jmr:
That's fine, but you'll have to [sign a CLA](https://github.com/google/s2geometry/blob/master/CONTRIBUTING.md). It's probably easier if I just fix them.

jmr:
It looks easier to sign the CLA than I thought. 

Do you want to send me a PR with some example fixes, then do everything when those look ok?

graetzer:
Yes, my collegue @danielhlarkin already submitted some of our patches, I will send some more when I get the time

PR #31, opened 2018-05-03
Silence some compiler warnings

graetzer:
I tried to silence some compiler warnings as mentioned in https://github.com/google/s2geometry/issues/13
I hope its not too much work to review this, but it does make building s2 a little nicer


jmr:
The class/struct changes are fine, as is the const return value one.

The unsigned ones, I'm not so sure about. @ericveach The code should follow the [Google style guide](https://google.github.io/styleguide/cppguide.html#Integer_Types), which doesn't use unsigned like this.

Maybe we just need to shut off the warning in CMakeLists.txt instead?

If you split this into multiple PRs, I can merge the one with the class/struct and const changes.

ericveach:
Regarding unsigned vs. int, I would prefer to just disable the warning.
The style guide used to be even more emphatic about this, but unsigned
types are discouraged except for bit manipulation and such.

I did a few quick searches over the Google code base, and it is about 50x
more common to use an "int" to iterate over vector indices than the next
most common type ("size_t"), which is then about 15x more common than
"unsigned" (i.e., "int" is about 750x more common than "unsigned").

jmr:
95646d1 and 363aa4b should fix the const int return value and struct/class warnings. 

a08bda6 Converted many loops to range-based for loops, and should remove some unsigned warnings.

If you're still seeing more unsigned warnings, please send a PR to change the compiler warning flags.

Issue #11, opened 2018-03-15
CMake newer version


hankim:
Hi,

I was trying to compile s2 python binding and had to update CMakefile a bit as I'm using newer version of CMake and some of directives are deprecated:

https://gist.github.com/hankim/d058ba1f26d9efd5aba296d6fa28e2e3

hankim:
btw, I have very little knowledge about CMake, so someone may find a better way of handling this.

hankim:
another issue is the dynamic library linking:
`@rpath/libs2.dylib (compatibility version 0.0.0, current version 0.0.0)`
and I had to manually change to:
`/usr/local/lib/libs2.dylib (compatibility version 0.0.0, current version 0.0.0)`
with the following command in Mac OS X:
`sudo install_name_tool -change @rpath/libs2.dylib /usr/local/lib/libs2.dylib /usr/local/share/python/_pywraps2.so`

I couldn't figure out where the link option is coming from.

PR #37, opened 2018-05-28
modernize cmake

### LINKED ISSUE #35: s2 adds `-std=c++11` uncoditonally
https://github.com/google/s2geometry/blame/master/CMakeLists.txt#L61

You should add this flag for clang/gcc to the private target definitions only. Or better use `set( CMAKE_CXX_STANDAD 11)` / comple features . It could look like: `target_compile_features(s2 PRIVATE cxx_std_11)`. Please use something newer than cmake 2.8.x. You have Blaze internally so don't force people to use antique cmake versions.

ObiWahn:
Please take a look at the current changes and let me know if they are something you like. If so I'll continue to work on my branch. This PR should eventually fix #35 #11 and maybe #6

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,9 @@
-cmake_minimum_required(VERSION 2.8.12)
+#TODO
+# - fix / modernize cmake
+# - lower cmake version a bit ~3.7 to support older systems
+# - fix abseil warnings or try to use SYSTEM / -isystem includes
+
+cmake_minimum_required(VERSION 3.10)
 project(s2-geometry)
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 include(CMakeDependentOption)
@@ -49,17 +54,17 @@ endif()
 find_package(OpenSSL REQUIRED)
 # pthreads isn't used directly, but this is still required for std::thread.
 find_package(Threads REQUIRED)
-find_package(SWIG)
-find_package(PythonInterp)
-find_package(PythonLibs)
 
 # Avoid "Policy CMP0042 is not set" warning on macOS.
 if (APPLE)
     set(CMAKE_MACOSX_RPATH TRUE)
 endif()
 
-add_definitions(-std=c++11)
+set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
+#TODO flags should be part of the compile interface - warnings should be only set on the project (example below)
+# target_compile_options(s2 PRIVATE $<$<CONFIG:Debug>:${-Wno-deprecated-declarations})>)
 if (WIN32)
     # Use unsigned characters
     add_definitions(-J)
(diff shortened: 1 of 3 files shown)


ObiWahn:
I signed it!

CMakeLists.txt
```diff
@@ -1,29 +1,38 @@
-cmake_minimum_required(VERSION 2.8.12)
+#TODO
+# - fix / modernize cmake
+# - lower cmake version a bit ~3.7 to support older systems
+# - fix abseil warnings or try to use SYSTEM / -isystem includes
+
+cmake_minimum_required(VERSION 3.8)
```
jmr: Separate PR for increasing cmake min version.  

What's the advantage of doing this? Removing version >= 3.8.0 checks?

CMakeLists.txt
```diff
@@ -1,29 +1,38 @@
...
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+
+#not used?
+#list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/third_party/cmake")
+#include(CheckCXXCompilerFlag)
+#include(FindPackageHandleStandardArgs)
+
+## set options
 include(CMakeDependentOption)
-include(CheckCXXCompilerFlag)
```
jmr: Separate PR to remove unneeded options.

CMakeLists.txt
```diff
@@ -1,29 +1,38 @@
...
 include(FeatureSummary)
-include(FindPackageHandleStandardArgs)
-
-list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/third_party/cmake")
 
 option(WITH_GFLAGS "Use gflags to change command line flags." OFF)
-add_feature_info(GFLAGS WITH_GFLAGS
-                 "allows changing command line flags.")
+add_feature_info(GFLAGS WITH_GFLAGS "allows changing command line flags.")
```
jmr: Don't reformat for no reason.

CMakeLists.txt
```diff
@@ -1,29 +1,38 @@
...
+# - fix abseil warnings or try to use SYSTEM / -isystem includes
+
+cmake_minimum_required(VERSION 3.8)
 project(s2-geometry)
+
+#produces compile_command.json
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+
+#not used?
+#list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/third_party/cmake")
```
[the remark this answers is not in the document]
ObiWahn: Oh good point! I have not tried all options yet.

CMakeLists.txt
ObiWahn: 2.8.x is really too old it does not really support targets (even though they became available in 2.8.11 iirc) and it does not have support for newer standards and language features. I think it is best to lower the version to something that still supports at least c++14, but something else might be ok. It could be better to use something that is available in the current Ubuntu release or in Travis containers. Eventually it is up to you! Please don't insist to support CentOS  6:)

CMakeLists.txt
ObiWahn: I am not sure if we will get on the same page here. I favor to rewrite the whole file on my branch and squash everything. I am willing to reason with you about every change, pass all tests and try to find the best solution. I do not want to create a meaningful commit with message for every line. That would be a too big time investment for me. Of course I will remove unrelated changes like in `coder.i`. And eventually smaller commits could be created when once the work is done.

CMakeLists.txt
ObiWahn: Is it ok to reformat everything to your style when I am done?

CMakeLists.txt
jmr: It looks like 3.8.0 is only a year old. That seems a bit new to require.

Maybe something in the 3.4-3.7 range? What's the oldest version that supports the features you want to use?

CMakeLists.txt
jmr: Sure. Is there a standard CMake indentation style?

jmr:
Closing stale PR with unresolved conflicts.

Issue #38, opened 2018-06-06
Error while building on windows


keyliaran:
C:\Users\shkil\Downloads\s2geometry-master\s2geometry-master\src\s2/third_party/absl/container/internal/layout.h(352): error C2971: std::tuple_element:

jmr:
What compiler version?  Can you paste more of the error log?

@danielhlarkin had the Windows version working at one point.

keyliaran:
@danielhlarkin @jmr Microsoft Visual Studio Community 2015v. 14.0.25421.03 Update 3
And here is the full log 
https://pastebin.com/ysq2MTuv

danielhlarkin:
Thanks for posting the full log. I'll look into it and see if I can figure out what's going on there.

jmr:
@danielhlarkin What version did you test with?

danielhlarkin:
@jmr Sorry, I have not had a chance to look into this yet. I don't actually have a Windows box of my own to play around with at the moment, but I should have one within the next few weeks. I believe the prior building/testing was done with MSVC 2015, but I do not have an exact version handy.

AndrewAtAvenza:
I have the same issue using MSVC 14.3. I found a post on stackoverflow that illuminated the issue (it's an MSVC compiler bug). I applied the suggested workaround in pull request #42 

I'm *not* suggesting it necessarily be merged. I'm mostly putting it up so @keyliaran might be able to continue on building the project, and maybe @danielhlarkin can evaluate a proper fix (that may or may not contain parts of what I posted).

AndrewAtAvenza:
I just installed Visual Studio 2017 (specifically MSVC 15.7.5) and tried building master without my changes and it compiles without problem. So whatever the bug was, they fixed it sometime in the 2017 time frame.

AluxRay:
@AndrewAtAvenza May I have a question?While building the project in Visual Sutdio 2017, there was a error "E2512 The parameter of the function test macro must be a simple identifier" in the line 535/617 of "attributes.h". I am looking forward to your help, thanks a lot.

AndrewAtAvenza:
Sorry, its been busy at work. I'll try it and take a look today and get back to you.

AndrewAtAvenza:
I just re-pulled and went through the (excruciating) process of configuring cmake. I finally spat out a 64-bit VS2017 solution and built the s2 project without any errors. 

What version of VS2017 are you running? I'm running 15.8.0 currently.

[...]

PR #42, opened 2018-07-19
Patch for issue #38

AndrewAtAvenza:
I don't necessarily suggest you integrate this, but this should fix the template error @keyliaran was having. I also have the same issue on MSVC 14.3.

The underlying issue is described here:

(https://stackoverflow.com/questions/50750642/msvc-a-variable-with-non-static-storage-duration-cannot-be-used-as-a-non-type-a)

After this, I can at least compile!


AndrewAtAvenza:
I have signed the Google CLA now.

jmr:
I'm guessing this is now irrelevant since these files are now taken from the released abseil, which probably works on Windows.  Update and re-open if that's not right.

PR #58, opened 2018-12-21
Removed unused aligned_malloc and aligned_free functions

bitweeder:
These were polluting the global namespace and conflicting with
identical symbols from other poorly-behaving libraries. Bugs are
being filed there, too, but s2’s case is particularly simple to fix.


jmr:
If you sign the CLA I can merge this.

jmr:
I thought I would have a better solution removing or greatly reducing base/port.h soon, but I'll merge this if you sign the CLA.

jmr:
Obsoleted by #213 .

PR #61, opened 2019-02-26
Update CMake to 3.5

bradhoov:
Reformat file for readbility and to emulate kitwares file layout/formatting

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,12 +1,20 @@
-cmake_minimum_required(VERSION 3.1)
+cmake_minimum_required(VERSION 3.5)
+
 project(s2-geometry)
-set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+
 include(CMakeDependentOption)
 include(CheckCXXCompilerFlag)
 include(FeatureSummary)
 include(FindPackageHandleStandardArgs)
-
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/third_party/cmake")
+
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -1,13 +1,21 @@
...
-
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/third_party/cmake")
 
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+
+# Avoid "Policy CMP0042 is not set" warning on macOS.
+if (APPLE)
+    set(CMAKE_MACOSX_RPATH TRUE)
+endif()
+
```
jmr: Delete extra blank line.

CMakeLists.txt
```diff
@@ -1,13 +1,21 @@
-cmake_minimum_required(VERSION 3.1)
+cmake_minimum_required(VERSION 3.5)
```
jmr: Split changing the min version and reorganizing into separate PRs. The reorganizing will be fine with a few adjustments. I'm still not clear on the reason for bumping the versions, but not against it.

CMakeLists.txt
```diff
@@ -53,11 +61,6 @@ find_package(SWIG)
 find_package(PythonInterp)
 find_package(PythonLibs)
 
-# Avoid "Policy CMP0042 is not set" warning on macOS.
-if (APPLE)
-    set(CMAKE_MACOSX_RPATH TRUE)
-endif()
-
 set(CMAKE_CXX_STANDARD 11)
```
jmr: Shouldn't these be moved with the other set()s above?

PR #71, opened 2019-05-09
Fix msvc warnings

dothebart:
These 3 commits will silence some verbose warnings we see with MSVC.

### DIFF
--- a/src/s2/base/port.h
+++ b/src/s2/base/port.h
@@ -58,6 +58,20 @@
 #define STDIN_FILENO 0
 #define STDOUT_FILENO 1
 #define STDERR_FILENO 2
+
+#ifdef S_IRGRP
+#undef S_IRGRP
+#endif
+#ifdef S_IRUSR
+#undef S_IRUSR
+#endif
+#ifdef S_IWGRP
+#undef S_IWGRP
+#endif
+#ifdef S_IWUSR
+#undef S_IWUSR
+#endif
+
 #define S_IRUSR 00400
 #define S_IWUSR 00200
 #define S_IXUSR 00100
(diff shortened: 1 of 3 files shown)


dothebart:
I signed it!

src/s2/s2region_coverer.h
```diff
@@ -28,6 +28,10 @@
 #include "s2/s2cell_id.h"
 #include "s2/s2cell_union.h"
 
+#ifdef _WIN32
+#pragma warning(disable : 4200)
```
jmr: Instead of disabling the warning, can we change the code, maybe deleting the copy-ctor/copy-assign it's complaining about, or using std::array with 0 size instead (not sure if that's legal, just throwing it out there as an idea)?

src/s2/base/port.h
```diff
@@ -58,6 +58,20 @@
 #define STDIN_FILENO 0
 #define STDOUT_FILENO 1
 #define STDERR_FILENO 2
+
+#ifdef S_IRGRP
```
jmr: How is this defined, and why is it sometimes defined and sometimes not with MSVC?

Prefer #ifndef ... #define ... #endif all as one block below.

src/s2/util/coding/coder.h
```diff
@@ -32,6 +32,10 @@
 #include "s2/util/coding/varint.h"
 #include "s2/util/endian/endian.h"
 
+#ifdef _WIN32
+#define memccpy _memccpy
```
jmr: Prefer to put everything in port.h.

jmr:
Closing due to inactivity. If this is still a a problem, open a new PR.

PR #96, opened 2019-12-05
Use local macro names to avoid conflicts with gflags.

gregcoombe:
The gflags macro names (`DEFINE_`) are also used by glog. If you have a setup that uses glog but not gflags, the `glog/logging.h` header will actually undefine these macros. This causes a very difficult-to-find error in S2. This PR changes the names of these macros to a library-specific name to avoid this issue. It follows the pattern of the logging macros.

### DIFF
--- a/src/s2/base/commandlineflags.h
+++ b/src/s2/base/commandlineflags.h
@@ -20,30 +20,42 @@
 
 #include <gflags/gflags.h>
 
+// Map local macros to GFlags macros.
+#define S2_DEFINE_bool  DEFINE_bool
+#define S2_DECLARE_bool  DECLARE_bool
+
+#define S2_DEFINE_double  DEFINE_double
+#define S2_DECLARE_double  DECLARE_double
+
+#define S2_DEFINE_int32  DEFINE_int32
+#define S2_DECLARE_int32  DECLARE_int32
+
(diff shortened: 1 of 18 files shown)


src/s2/base/commandlineflags.h
```diff
@@ -20,30 +20,42 @@
 
 #include <gflags/gflags.h>
 
+// Map local macros to GFlags macros.
```
jmr, line 23: Expand on this comment for why you want to do this.

jmr:
It seems like this is a glog bug, but we probably just have to work around it. Can you file a bug / send a PR there?

gregcoombe:
> It seems like this is a glog bug, but we probably just have to work around it. Can you file a bug / send a PR there?

So I'm not sure that this is really a glog bug, or that they would be able to resolve it. It's more of an issue with how the different libraries are used within my system, which interacts badly with the fact that S2 uses the same names (in a global namespace) as gflags for macros.
The reason that I thought that this would be a good fix is that using library-specific macros is a better practice overall (S2 already does this for glog) and it solves my immediate problem. So I don't really think of this as a work-around! I think of it as a tiny step forward in the design of the library.

jievince:
I am using V9.0 and this problem also took me a long time to debug

PR #92, opened 2019-11-15
Remove "using std::string" from port.h.

gregcoombe:
Exposing the "using" directive in the port header will cause it
to be leaked into any codebase that uses this library. While it
is possible to just fix up the headers and use the directive
only in the source files, this change uses std::string
everywhere.

PR #97, opened 2019-12-05
Replace custom integral types with stdint.h types.

gregcoombe:
The `base/integral_types.h` file defines a number of fixed-width integral types by typedef-ing them to (theoretically) variable-width types. This PR replaces the local types with the common fixed-width types throughout the codebase, and removes this port file.


jmr:
Can you #include <cstdint> and use std::... instead?

gregcoombe:
Are you suggesting to just rewrite `base/integral_types.h` to use `std::...` types but still continue to include it? My thought was that this simplified the code by removing an unnecessary header. But it does make it a little bit larger diff.

The other reason for using the built-in types is that it reduces the number of custom types that leak into codebases. These types are in the global namespace, not contained within an S2 namespace, so they will interfere with any other types defined elsewhere. Perhaps this is minor concern, since it would strange to have `uint8` defined as anything else except a 8-bit int, but it seemed to me to be worth the effort of removing that potential source of error.

jmr:
> Are you suggesting to just rewrite base/integral_types.h to use std::... types but still continue to include it?

No, just to use `std::int32_t` instead of plain `int32_t`.

jmr:
I will close and re-open this in hope of getting travis-ci to notice the .travis.yml changes. The current CMakeLists.txt is incompatible with the old .travis.yml.

src/s2/s1angle_test.cc
```diff
@@ -71,16 +70,16 @@ TEST(S1Angle, E6E7RepresentationsUnsigned) {
   // Check that unsigned E6/E7 representations work as expected.
   EXPECT_DOUBLE_EQ(
       S1Angle::Degrees(60).radians(),
-      S1Angle::UnsignedE6(static_cast<uint32>(60000000)).radians());
+      S1Angle::UnsignedE6(static_cast<std::uint32_t>(60000000)).radians());
```
jmr, line 73: Let's do the same `using std::uint32_t;` thing in the .cc files that #92 does with `std::string`.

jmr:
Tests are failing.

jmr:
This was done in 755fe37c9fda541b7b9bcb725b03803d2e807dac.

PR #112, opened 2020-07-07
Add S2CellId::levelpos() method 

r-barnes:


src/s2/s2cell_id_test.cc
```diff
@@ -81,6 +81,7 @@ TEST(S2CellId, ParentChildRelationships) {
   EXPECT_TRUE(id.is_valid());
   EXPECT_EQ(3, id.face());
   EXPECT_EQ(0x12345700, id.pos());
+  EXPECT_EQ(0x91a2b, id.levelpos());
```
jmr, line 84: Could you check if this is the same as [`S2CellId::distance_from_begin`](https://github.com/google/s2geometry/blob/master/src/s2/s2cell_id.cc#L131)?

r-barnes: @jrm: The mathematics work out the same internally, however [`S2CellId::distance_from_begin`](https://github.com/google/s2geometry/blob/master/src/s2/s2cell_id.cc#L131) returns a `int64_t`. It seems as though it should return a `uint64_t`.

`levelpos()` also is analogous to the existing [`S2CellId::pos()`](https://github.com/google/s2geometry/blob/master/src/s2/s2cell_id.h#L206-L209).

A few potential resolutions:
* `S2CellId::distance_from_begin` should probably return `uint64_t` regardless.
* `levelpos()` could alias to `distance_from_begin()` with test case for equality.
* Or, the documentation for `distance_from_begin()` could be updated to make it clearer that it is an analogous operation to `pos()`.

jmr: > It seems as though it should return a uint64_t.

s2geometry uses the [Google C++ style guide](https://google.github.io/styleguide/cppguide.html#Integer_Types), which avoids unsigned due to surprises like `1U - 2 > 0`.

> Or, the documentation for distance_from_begin() could be updated

I'll do this.

Issue #114, opened 2020-07-07
Compilation warning for test suite


r-barnes:
Compiling the test suite gives me this warning:
```
[26/319] Building CXX object CMakeFiles/s2builderutil_lax_polygon_layer_test.dir/src/s2/s2builderutil_lax_polygon_layer_test.cc.o
../src/s2/s2builderutil_lax_polygon_layer_test.cc: In function ‘std::string {anonymous}::ToString(DegenerateBoundaries)’:
../src/s2/s2builderutil_lax_polygon_layer_test.cc:56:1: warning: control reaches end of non-void function [-Wreturn-type]
   56 | }
      | ^
[319/319] Linking CXX executable s2lax_polygon_shape_test
```

This is because there is a switch here to convert enum values to strings, but the switch has no default option (for cases in which the enum has a value the switch doesn't pick up on).

r-barnes:
(Author/Contributor info is in #112 )

PR #115, opened 2020-07-07
Fix compilation warning in s2builderutil_lax_polygon_layer_test.cc

r-barnes:
Fixes the issue described in #114

### DIFF
--- a/src/s2/s2builderutil_lax_polygon_layer_test.cc
+++ b/src/s2/s2builderutil_lax_polygon_layer_test.cc
@@ -21,6 +21,7 @@
 #include <map>
 #include <memory>
 #include <set>
+#include <stdexcept>
 #include <string>
 #include <gtest/gtest.h>
 #include "s2/base/integral_types.h"
@@ -53,6 +54,7 @@ string ToString(DegenerateBoundaries degenerate_boundaries) {
     case DegenerateBoundaries::DISCARD_SHELLS: return "DISCARD_SHELLS";
     case DegenerateBoundaries::KEEP: return "KEEP";
   }
+  S2_LOG(FATAL) << "Unrecognized DegenerateBoundaries type!";
 }
 
 void TestLaxPolygon(string_view input_str, string_view expected_str,


src/s2/s2builderutil_lax_polygon_layer_test.cc
```diff
@@ -52,6 +53,8 @@ string ToString(DegenerateBoundaries degenerate_boundaries) {
     case DegenerateBoundaries::DISCARD_HOLES: return "DISCARD_HOLES";
     case DegenerateBoundaries::DISCARD_SHELLS: return "DISCARD_SHELLS";
     case DegenerateBoundaries::KEEP: return "KEEP";
+    default:
+      throw std::runtime_error("Unrecognized DegenerateBoundaries value!");
   }
 }
```
jmr, line 59: `s2geometry` doesn't use exceptions, so `return "UNKNOWN";` or `LOG(FATAL) << ...` *after* the switch would be better, but I'm really wondering why you're getting this warning when the cases are exhaustive.

r-barnes: I'm also a little mystified as to why the compiler wouldn't detect this, but I do get this warning with an unaltered configuration using G++ 9.3.0. I've updated the commit to use `LOG`, as you suggest.

r-barnes:
Not sure if I should have resolved the above conversation. My reply:
> I'm also a little mystified as to why the compiler wouldn't detect this, but I do get this warning with an unaltered configuration using G++ 9.3.0. I've updated the commit to use LOG, as you suggest.

src/s2/s2builderutil_lax_polygon_layer_test.cc
```diff
@@ -53,6 +54,7 @@ string ToString(DegenerateBoundaries degenerate_boundaries) {
     case DegenerateBoundaries::DISCARD_SHELLS: return "DISCARD_SHELLS";
     case DegenerateBoundaries::KEEP: return "KEEP";
   }
+  S2_LOG(FATAL) << "Unrecognized DegenerateBoundaries type!";
```
jmr, line 57: How about
```
S2_LOG(FATAL) << "Unknown DegenerateBoundaries value: "
              << static_cast<int>(degenerate_boundaries);
```
?

I can also just fix it if you want.

src/s2/s2builderutil_lax_polygon_layer_test.cc
```diff
@@ -21,6 +21,7 @@
 #include <map>
 #include <memory>
 #include <set>
+#include <stdexcept>
```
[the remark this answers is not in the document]
jmr: This was never fixed, so merging #136 instead.

PR #128, opened 2020-08-28
Require Python3

jmr:
This should help solve problems where s2's SWIG module was build with
Python2, but users try to use Python3.

Python2 has been sunsetted as of 2020-01-01.

Increases required CMake version from 3.5 (released 2016-03) to 3.12
(released 2018-07).


jmr:
travis-ci doesn't have a new enough cmake, so I need to fix that first -- #129.

PR #144, opened 2020-09-19
fixed arguments order in line 51

alas:
small fix


rsned:
https://github.com/google/s2geometry/blob/master/src/s2/s1interval.h#L53

S1Interval(double lo, double hi);

The complement of (lo, hi), is (hi, lo)

alas:
thank you!

PR #145, opened 2020-09-19
Expose S2CellId::AppendAllNeighbors in Python (as GetAllNeighbors)

micolous:
This exposes `S2CellId::AppendAllNeighbors` to Python code as `S2CellId.GetAllNeighbors`:

```python
>>> import pywraps2 as s2
>>> p = s2.S2CellId.FromToken('6aa759')
>>> ','.join(_.ToToken() for _ in p.GetAllNeighbors(p.level()))
'2ab353,2ab34b,2ab34d,6aa75b,6aa757,6aa75f,6aa751,6aa75d'
```

I renamed this because "Append" doesn't make sense – SWIG is configured to convert passing a (mutable) output `std::vector` by reference into returning an immutable `tuple`.

This was exposed in the API before 13ab3fc4 removed all members by default.

### DIFF
--- a/src/python/pywraps2_test.py
+++ b/src/python/pywraps2_test.py
@@ -63,6 +63,19 @@ class PyWrapS2TestCase(unittest.TestCase):
     neighbors = cell.GetEdgeNeighbors()
     self.assertEqual(neighbors, expected_neighbors)
 
+  def testS2CellIdGetAllNeighborsIsWrappedCorrectly(self):
+    cell = s2.S2CellId(0x6aa7590000000000)
+    expected_neighbors = (s2.S2CellId(0x2ab3530000000000),
+                          s2.S2CellId(0x2ab34b0000000000),
+                          s2.S2CellId(0x2ab34d0000000000),
+                          s2.S2CellId(0x6aa75b0000000000),
+                          s2.S2CellId(0x6aa7570000000000),
+                          s2.S2CellId(0x6aa75f0000000000),
+                          s2.S2CellId(0x6aa7510000000000),
+                          s2.S2CellId(0x6aa75d0000000000))
+    neighbors = cell.GetAllNeighbors(cell.level())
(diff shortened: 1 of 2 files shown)


src/python/pywraps2_test.py
```diff
@@ -63,6 +63,19 @@ def testS2CellIdGetEdgeNeighborsIsWrappedCorrectly(self):
     neighbors = cell.GetEdgeNeighbors()
     self.assertEqual(neighbors, expected_neighbors)
 
+  def testS2CellIdGetAllNeighborsIsWrappedCorrectly(self):
+    cell = s2.S2CellId(0x6aa7590000000000)
+    expected_neighbors = (s2.S2CellId(0x2ab3530000000000),
```
jmr, line 68: Use a list, not a tuple, for consistency with above and with what it actually returns.

jmr (changes requested):

src/python/pywraps2_test.py
micolous: It actually returns a tuple:

```python
>>> import pywraps2 as s2
>>> p = s2.S2CellId.FromToken('6aa759')
>>> type(p.GetAllNeighbors(p.level()))
<class 'tuple'>
```

I haven't configured anything special in SWIG to do this – [this is the default behaviour](https://stackoverflow.com/questions/52960876/tuple-is-returned-by-python-wrapper-generated-by-swig-for-a-c-vector).

src/python/s2_common.i
```diff
@@ -311,6 +311,8 @@ class S2Point {
 %unignore S2CellId::FromToken(const std::string&);
 %unignore S2CellId::GetCenterSiTi(int*, int*) const;
 %unignore S2CellId::GetEdgeNeighbors;
+%unignore S2CellId::AppendAllNeighbors;
```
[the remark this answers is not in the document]
micolous: Done.

PR #150, opened 2020-10-14
removed declaration of function without implementation

alas:
this function doesn't have an implementation and is it not used anywhere


jmr:
Actually this one should be implemented. It was just a mistake that it was omitted.

jmr:
#152 added a definition for this.

PR #160, opened 2020-12-01
Expand Python Bindings to include S2RegionTermIndexer

### LINKED ISSUE #84: Python bindings do not include S2RegionTermIndexer
`S2RegionTermIndexer` contains useful logic that would be great to have in the Python bindings.

Is there a particular reason this class isn't included? I'm unfortunately out of my depth in trying to modify the bindings, and instead planning to just reimplement some of it in the python code directly. Thanks.

rcoup:
* Make it possible to follow the _Indexing for Search_ and _Querying for Search_ sections in the [Quickstart guide](https://s2geometry.io/devguide/cpp/quickstart#indexing-for-search) from Python.
* Primarily expand the Python bindings to include `S2RegionTermIndexer`
* And related changes for testing
* Port the `term_index.cc` example to Python as well for good measure.
* Add some `repr()` methods

Fixes #84

### DIFF
--- a/doc/examples/term_index.py
+++ b/doc/examples/term_index.py
@@ -1,0 +1,101 @@
+"""
+This example shows how to add spatial data to an information retrieval
+system.  Such systems work by converting documents into a collection of
+"index terms" (e.g., representing words or phrases), and then building an
+"inverted index" that maps each term to a list of documents (and document
+positions) where that term occurs.
+
+This example shows how to convert spatial data into index terms, which can
+then be indexed along with the other document information.
+
+This is a port of the C++ term_index.cc example for the Python API.
+"""
+import argparse
+from collections import defaultdict
+
+import pywraps2 as s2
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        description=(
+            "This example shows how to convert spatial data into index terms, "
+            "which can then be indexed along with the other document "
+            "information."
+        )
+    )
+    parser.add_argument(
+        '--num_documents', type=int, default=10000, help="Number of documents"
+    )
+    parser.add_argument(
+        '--num_queries', type=int, default=10000, help="Number of queries"
+    )
+    parser.add_argument(
+        '--query_radius_km', type=float, default=100,
+        help="Query radius in kilometers"
+    )
+
+    args = parser.parse_args()
+
+    # A prefix added to spatial terms to distinguish them from other index terms
+    # (e.g. representing words or phrases).
+    PREFIX = "s2:"
+
+    # Create a set of "documents" to be indexed.  Each document consists of a
+    # single point.  (You can easily substitute any S2Region type here, or even
+    # index a mixture of region types using S2Region.  Other
+    # region types include polygons, polylines, rectangles, discs, buffered
+    # geometry, etc.)
+    documents = []
+    for i in range(args.num_documents):
+        documents.append(s2.S2Testing.RandomPoint())
+
+    # We use a dict as our inverted index.  The key is an index term, and
+    # the value is the set of "document ids" where this index term is present.
+    index = defaultdict(set)
+
+    # Create an indexer suitable for an index that contains points only.
+    # (You may also want to adjust min_level() or max_level() if you plan
+    # on querying very large or very small regions.)
+    indexer = s2.S2RegionTermIndexer()
+    indexer.set_index_contains_points_only(True)
+
+    # Add the documents to the index.
+    for docid, index_region in enumerate(documents):
+        for term in indexer.GetIndexTerms(index_region, PREFIX):
+            index[term].add(docid)
+
+    # Convert the query radius to an angle representation.
+    radius = s2.S1Angle.Radians(s2.S2Earth.KmToRadians(args.query_radius_km))
+
+    # Count the number of documents (points) found in all queries.
+    num_found = 0
+    for i in range(args.num_queries):
+        # Choose a random center for query.
+        query_region = s2.S2Cap(s2.S2Testing.RandomPoint(), radius)
+
+        # Convert the query region to a set of terms, and compute the union of
+        # the document ids associated with those terms.  (An actual information
+        # retrieval system would do something more sophisticated.)
+        candidates = set()
+        for term in indexer.GetQueryTerms(query_region, PREFIX):
+            candidates |= index[term]
+
+        # "candidates" now contains all documents that intersect the query
+        # region, along with some documents that nearly intersect it.  We can
+        # prune the results by retrieving the original "document" and checking
+        # the distance more precisely.
+        result = []
+        for docid in candidates:
+            if query_region.Contains(documents[docid]):
+                result.append(docid)
+
+        # Now do something with the results (in this example we just count
+        # them).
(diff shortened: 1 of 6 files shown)


rcoup:
@googlebot I signed it!

src/python/pywraps2_test.py
```diff
@@ -606,5 +632,166 @@ def testS2CellIdToFromFaceIJ(self):
...
+        if covering.Intersects(coverings[j]):
+          expected.add(j)
+      
+      for term in terms:
+        actual |= index[term]
+
+      self.assertEqual(expected, actual)
+      query_terms += len(terms)
+
+      # print("Index terms/doc: %0.2f, Query terms/doc: %0.2f" % (float(index_terms) / ITERATIONS, float(query_terms) / ITERATIONS))
```
jmr, line 718: might as well print it since the C++ version does

src/python/s2_common.i
```diff
@@ -34,6 +36,16 @@
 %apply std::vector<S2CellId> *OUTPUT {std::vector<S2CellId> *interior};
 %apply std::vector<S2CellId> *OUTPUT {std::vector<S2CellId> *output};
 
+%typemap(in) absl::string_view {
+  if (PyUnicode_Check($input)) {
+    $1 = absl::string_view(PyUnicode_AsUTF8($input));
+  } else {
+    SWIG_exception(SWIG_TypeError, "string expected");
```
jmr, line 43: a string_view may also be pointing at non-utf-8 data, but I guess that's not happening. How do we add that if we need to?

src/python/s2_common.i
```diff
@@ -127,8 +139,17 @@ std::vector<S2Polyline *> *out(std::vector<S2Polyline *> temp) {
   }
 }
 
+%inline %{
+  // mostly useful for testing
+  static PyObject *S2Point_FromRaw(double x, double y, double z) {
```
jmr, line 144: Can this be a constructor?

src/python/s2_common.i
```diff
@@ -127,8 +139,17 @@ std::vector<S2Polyline *> *out(std::vector<S2Polyline *> temp) {
...
+    return SWIG_NewPointerObj(new S2Point(x, y, z), SWIGTYPE_p_S2Point,
+                              SWIG_POINTER_OWN);
+  }
+%}
+
 // We provide our own definition of S2Point, because the real one is too
 // difficult to wrap correctly.
+
```
jmr, line 152: delete this added empty line

src/python/s2_common.i
```diff
@@ -266,6 +334,8 @@ class S2Point {
 %unignore S2Cap::Empty;
 %unignore S2Cap::Encode;
 %unignore S2Cap::Expanded;
+%unignore S2Cap::FromCenterHeight(const S2Point&, double);
+%unignore S2Cap::FromCenterArea(const S2Point&, double);
```
jmr, line 338: Keep these in alphabetical order.

src/python/s2_common.i
```diff
@@ -362,6 +433,41 @@ class S2Point {
 %unignore S2CellUnion::cell_ids;
 %unignore S2CellUnion::empty;
 %unignore S2CellUnion::num_cells;
+%unignore S2Earth;
```
jmr, line 436: alphabetical order, use your editor to sort them

jmr (changes requested):
Thanks for the contribution!

doc/examples/term_index.py
```diff
@@ -0,0 +1,102 @@
...
+This is a port of the C++ term_index.cc example for the Python API.
+"""
+import argparse
+import random
+from collections import defaultdict
+
+import pywraps2 as s2
+
+
+def RandomPoint():
```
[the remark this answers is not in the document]
rcoup: The S2Testing C++ class is conditionally built, and not with #defines, so it didn't seem trivial to conditionally build Python bindings to it. For the time I had, seemed easier to reimplement it.

src/python/s2_common.i
rcoup: The string in this case is coming from python though, and [SWIG's default behaviour is to map std::string ↔︎ String](http://www.swig.org/Doc3.0/Python.html#Python_nn77) in Py3.

Could use `SWIG_PYTHON_STRICT_BYTE_CHAR` and force all string encoding to happen in Python. Not sure if that solves a real use-case though?

src/python/s2_common.i
rcoup: I had a go at that initially and ran into issues because of the [`S2Point` redefinition](https://github.com/google/s2geometry/blob/master/src/python/s2_common.i#L130-L137) happening. Suggestions welcome.

src/python/pywraps2_test.py
rcoup: unittest has terrible output-control cf pytest/etc, but I guess can suppress via `-b`

src/python/s2_common.i
rcoup: I guess I could easily convert Python Bytes without encoding to `std::string`, but not sure what would happen on the way back out, considering SWIG's default behaviour.

src/python/pywraps2_test.py
```diff
@@ -14,12 +14,38 @@
...
-
-
+import random
 import unittest
+from collections import defaultdict
+
 import pywraps2 as s2
 
 
+class S2Testing(object):
```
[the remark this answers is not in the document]
rcoup: I moved it to `pywraps2.S2Testing`

rcoup:
@jmr ready for another look

src/python/s2_common.i
jmr: Could you just add a comment that that's what's happening? Maybe we'll figure out how to clean it up later.

rcoup: done

rcoup:
@googlebot I fixed it

rcoup:
@jmr I think this is good to go now.

src/python/s2_common.i
```diff
@@ -597,3 +723,72 @@ USE_COMPARISON_FOR_LT_AND_GT(S1Angle)
 USE_EQUALS_FN_FOR_EQ_AND_NE(S2Loop)
 USE_EQUALS_FN_FOR_EQ_AND_NE(S2Polygon)
 USE_EQUALS_FN_FOR_EQ_AND_NE(S2Polyline)
+
```
jmr, line 726: Can this section be a separate PR? Everything else looks good, but now I have to go learn about repr()

src/python/s2_common.i
```diff
@@ -597,3 +723,72 @@ USE_COMPARISON_FOR_LT_AND_GT(S1Angle)
...
+// repr() methods for geodetic/cell classes
+
+%extend S2CellId {
+  %pythoncode %{
+    def __repr__(self):
+      return '<S2CellId: %s>' % self.ToToken()
```
jmr, line 732: Should this eval back to the same thing? If not, ToString() is more readable and standard for debugging in c++.

rcoup:
Done, will do a new PR for repr

PR #161, opened 2020-12-05
CMake: add version information to the library

PureTryOut:
On Linux systems, libraries need to be installed with attached version
information so other packages can depend on specific versions and make
sure they don't compile against a version they're not compatible with

This change installs libs2.so.0.10.0 with appropriate
symlinks as libs2.so.0 and libs2.so.

Do note that for new future versions, the version in CMakeLists.txt will have
to be updated as well

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,7 @@
 cmake_minimum_required(VERSION 3.12)
 
-project(s2-geometry)
+project(s2-geometry
+	VERSION 0.10.0)
 
 include(CMakeDependentOption)
 include(CheckCXXCompilerFlag)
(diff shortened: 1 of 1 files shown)


PureTryOut:
@googlebot I signed it!

CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 cmake_minimum_required(VERSION 3.12)
 
-project(s2-geometry)
+project(s2-geometry
+	VERSION 0.10.0)
```
jmr, line 4: What are we allowed to put here? Can this be 0.10.0.dev or 0.10.0.alpha?

Also note that we don't actually guarantee any API stability currently, but having some version number is better than none.

PureTryOut: It accepts a "tweak" version, so e.g. `0.10.0.1`, but it has to be a number, it can't  be any other type of character.

Documentation: https://cmake.org/cmake/help/latest/command/project.html#command:project

jmr: Is it good practice to put version numbers for the next release in a development version? It seems like it could lead to grief, since the API could change before the release.

Maybe put a comment to that effect.

jmr: Ok, let's see if anyone complains this way.

PR #170, opened 2021-02-08
Make python and testing optional

wegylexy:


CMakeLists.txt
```diff
@@ -88,7 +88,10 @@ endif()
 # OPENSSL_ROOT_DIR=/usr/local/opt/openssl cmake ..
 include_directories(
     ${GFLAGS_INCLUDE_DIRS} ${GLOG_INCLUDE_DIRS} ${OPENSSL_INCLUDE_DIR}
-    ${Python3_INCLUDE_DIRS})
+)
+if (${Python3_FOUND})
```
jmr, line 92: I think it would be better to make this more explicit, like `option(WITH_PYTHON, ...)`. Otherwise, we just silently don't get python, which could be confusing.

jmr:
`BUILD_TESTS` and `WITH_PYTHON` can do this.

PR #169, opened 2021-01-06
Remove internal abseil

jmr:
Require pointing at an abseil source tree with cmake -DABSL_SOURCE_DIR=.

Fixes #50, #153, #168.

PR #174, opened 2021-03-12
Remove use of std::allocator::rebind

### LINKED ISSUE #173: not compatible with c++20
std::allocator::rebind was deprecated in c++17, and removed in c++20. It should be replaced with std::allocator_traits::rebind_alloc. The use of allocator rebind means when compiling with GCC10 and -std=c++20, compilation fails with something like

```
In file included from /usr/local/include/s2/s2shape_index.h:44,
                 from /usr/local/include/s2/mutable_s2shape_index.h:37,
                 from /usr/local/include/s2/s2polygon.h:29,
                 from ...:
/usr/local/include/s2/util/gtl/compact_array.h: In instantiation of 'class gtl::compact_array_base<S2ClippedShape, std::allocator<S2ClippedShape> >':
/usr/local/include/s2/util/gtl/compact_array.h:520:7:   required from 'class gtl::compact_array<S2ClippedShape>'
/usr/local/include/s2/s2shape_index.h:157:21:   required from here
/usr/local/include/s2/util/gtl/compact_array.h:148:49: error: no class template named 'rebind' in 'class std::allocator<S2ClippedShape>'
  148 |   typedef typename A::template rebind<T>::other value_allocator_type;
      |                                                 ^~~~~~~~~~~~~~~~~~~~
In file included from /usr/local/include/s2/util/gtl/dense_hash_set.h:128,
                 from /usr/local/include/s2/sequence_lexicon.h:26,
                 from /usr/local/include/s2/id_set_lexicon.h:26,
                 from /usr/local/include/s2/s2builder.h:30,
                 from /usr/local/include/s2/s2boolean_operation.h:24,
                 from /usr/local/include/s2/s2polygon.h:31,
```

jmr:
This is deprecated in C++17 and has been removed in C++20.

Use std::allocator_traits::rebind_alloc instead.

https://en.cppreference.com/w/cpp/memory/allocator
https://en.cppreference.com/w/cpp/memory/allocator_traits

Fixes #173.

### DIFF
--- a/src/s2/util/gtl/compact_array.h
+++ b/src/s2/util/gtl/compact_array.h
@@ -145,7 +145,8 @@ class compact_array_base {
     return const_cast<compact_array_base<T, A>*>(this)->Array();
   }
 
-  typedef typename A::template rebind<T>::other value_allocator_type;
+  using value_allocator_type =
+      typename std::allocator_traits<A>::template rebind_alloc<T>;
 
  public:
   typedef T                                     value_type;
--- a/src/s2/util/gtl/densehashtable.h
+++ b/src/s2/util/gtl/densehashtable.h
@@ -182,7 +182,8 @@ struct dense_hashtable_const_iterator;
 template <class V, class K, class HF, class ExK, class SetK, class EqK, class A>
 struct dense_hashtable_iterator {
  private:
-  typedef typename A::template rebind<V>::other value_alloc_type;
+  using value_alloc_type =
+      typename std::allocator_traits<A>::template rebind_alloc<V>;
 
  public:
   typedef dense_hashtable_iterator<V, K, HF, ExK, SetK, EqK, A>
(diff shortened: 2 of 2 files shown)


nbronson:
Thanks for the quick response. The bug report didn't include all issues blocking C++20 build, unfortunately, as std::allocator has also removed the pointer, const_pointer, reference, and const_reference aliases. These are used from src/s2/util/gtl/densehashtable.h and also a few places in third_party/absl (fixed_array.h, inlined_vector.h), so a complete fix will require an abseil upgrade.

jmr:
For the ones in absl, can you check https://github.com/abseil/abseil-cpp and file any bugs there?  We'll be switching to that as soon as I add it to the travis-ci config (just in time for travis-ci.org to be turned off).  If you could also look at #169 and see if that works for you, that would be great, too.  Do you already use absl?

Is the replacement for `std::allocator::pointer` `allocator_traits<...>::pointer`, etc?

nbronson:
I'll check abseil. We don't use it directly, only indirectly.

`pointer` and `const_pointer` should be taken from `allocator_traits`. `reference` and `const_reference` are not in `allocator_traits`. The C++ standard requires that containers define them to be `value_type&` and `const value_type&` (section 26.2.1 in the N4713 draft of the C++17 standard, section 22.2.1 in the N4861 draft of the C++20 standard).

PR #184, opened 2021-04-18
python/CMakeLists.txt: Fix install destination

podsvirov:
Added `CMAKE_INSTALL_SITELIBDIR` cached variable.
Default value is set to `Python3_SITELIB` (absolute path),
but can be changed at configuration time.
An absolute path is not acceptable when creating an installation package.


src/python/CMakeLists.txt
```diff
@@ -15,7 +15,10 @@ add_test(NAME pywraps2_test COMMAND
 set_property(TEST pywraps2_test PROPERTY ENVIRONMENT
              "PYTHONPATH=$ENV{PYTHONPATH}:${PROJECT_BINARY_DIR}/python")
 
+set(CMAKE_INSTALL_SITELIBDIR "${Python3_SITELIB}"
```
jmr, line 18: Calling this `CMAKE_` seems a bit strange. I would assume that's only for variables defined by CMake itself.

Can you point me at some CMake Python documentation or a canonical example? It seems no one has used this variable name before.

https://www.google.com/search?q=%22CMAKE_INSTALL_SITELIBDIR%22
https://github.com/search?q=CMAKE_INSTALL_SITELIBDIR

jmr: Also, what problem are you trying to solve? Use of this variable should also be documented somewhere.

podsvirov: I'm trying to do something like there: https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html.

podsvirov: @jmr, please review #186 (this should clarify my motivation).

jmr:
Does #186 make this unnecessary?

podsvirov:
These changes are not directly related to the #186. These changes solve the problem of hard-coded absolute installation path for the python module. Absolute install paths add additional problems when creating packages for distribution (in this case, it becomes necessary to install to a path other than the default).

src/python/CMakeLists.txt
jmr: It seems like there should be a relevant directory somewhere here:
https://cmake.org/cmake/help/latest/module/FindPython3.html

How do other swig/python/windows/cmake projects do it?

jmr:
What do other Python people think?  @andyneff

PR #88, opened 2019-10-24
Add MINGW to port.h

rubak:
These minor changes makes the library compile with MINGW on my machine.
Some changes may be obsolete if PR #58 is merged at some point.

PR #185, opened 2021-04-18
Add MinGW port

podsvirov:
Tested for 32bit and 64bit targets via [MSYS2](https://www.msys2.org/) distro on `Windows`.

Similar PR #88.

### DIFF
--- a/src/s2/base/port.h
+++ b/src/s2/base/port.h
@@ -290,11 +290,11 @@ inline void sized_delete_array(void *ptr, size_t size) {
 
 #endif
 
-// defines __BYTE_ORDER for MSVC
-#ifdef _MSC_VER
+// defines __BYTE_ORDER
+#ifdef _WIN32
 #define __BYTE_ORDER __LITTLE_ENDIAN
 #define IS_LITTLE_ENDIAN
-#else
+#else // _WIN32
 
 // define the macros IS_LITTLE_ENDIAN or IS_BIG_ENDIAN
 // using the above endian definitions from endian.h if
@@ -317,7 +317,7 @@ inline void sized_delete_array(void *ptr, size_t size) {
 #endif
 
 #endif  // __BYTE_ORDER
-#endif  // _MSC_VER
+#endif  // _WIN32
 
 // byte swap functions (bswap_16, bswap_32, bswap_64).
 
@@ -602,11 +602,11 @@ std::ostream &operator<<(std::ostream &out, const pthread_t &thread_id);
 #endif  // _MSC_VER
 
 // random, srandom
-#ifdef _MSC_VER
+#ifdef _WIN32
 // You say tomato, I say toma
 inline int random() { return rand(); }
 inline void srandom(unsigned int seed) { srand(seed); }
-#endif  // _MSC_VER
+#endif  // _WIN32
 
 // bcopy, bzero
 #ifdef _MSC_VER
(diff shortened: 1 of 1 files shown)


src/s2/base/port.h
```diff
@@ -290,8 +290,8 @@ inline void sized_delete_array(void *ptr, size_t size) {
 
 #endif
 
-// defines __BYTE_ORDER for MSVC
-#ifdef _MSC_VER
+// defines __BYTE_ORDER for _WIN32
```
jmr, line 293: There are a bunch of other places that use `_MSC_VER`:
https://github.com/google/s2geometry/search?q=_msc_ver

None of these need to change?

src/s2/base/port.h
```diff
@@ -888,7 +888,7 @@ inline void UnalignedCopy64(const void *src, void *dst) {
        defined(__NVCC__)) &&   \
       !defined(SWIG)) ||                                                \
      ((__GNUC__ >= 3 || defined(__clang__)) && defined(__ANDROID__)) || \
-     defined(__ASYLO__))
+     defined(__ASYLO__)) && !defined(__MINGW32__)
```
jmr, line 891: Why does this need to be added? Can you add a comment above?

src/s2/base/port.h
jmr: Does `__MINGW32__` imply `_WIN32`?

jmr (changes requested):

src/s2/base/port.h
podsvirov: In my experience, yes.

podsvirov:
Reworked some edits to have fewer questions.
@jmr, please, review again.

src/s2/base/port.h
```diff
@@ -290,8 +290,8 @@ inline void sized_delete_array(void *ptr, size_t size) {
 
 #endif
 
-// defines __BYTE_ORDER for MSVC
-#ifdef _MSC_VER
+// defines __BYTE_ORDER for MSVC or MINGW
+#if defined(_MSC_VER) || defined(__MINGW32__)
```
jmr, line 294: `_MSC_VER` was probably a mistake. If `_WIN32` works for both windows and MinGW, let's use that.

jmr (changes requested):

src/s2/base/port.h
```diff
@@ -602,11 +602,11 @@ std::ostream &operator<<(std::ostream &out, const pthread_t &thread_id);
 #endif  // _MSC_VER
 
 // random, srandom
-#ifdef _MSC_VER
+#if defined(_MSC_VER) || defined(__MINGW32__)
```
jmr, line 605: Everything that's actually related to windows rather than MSVC should be switched to `_MSC_VER`.

src/s2/base/port.h
```diff
@@ -883,7 +883,7 @@ inline void UnalignedCopy64(const void *src, void *dst) {
 
 // __ASYLO__ platform uses newlib without an underlying OS, which provides
 // memalign, but not posix_memalign.
-#if defined(__cplusplus) &&                                             \
+#if !defined(__MINGW32__) && defined(__cplusplus) &&                    \
     (((defined(__GNUC__) || defined(__APPLE__) ||                  \
        defined(__NVCC__)) &&   \
       !defined(SWIG)) ||                                                \
```
jmr, line 889: Move !defined(__MINGW32__) to after !defined(SWIG). This keeps the exceptions together.

jmr (changes requested):

jmr:
Thanks!

PR #186, opened 2021-04-18
CMake: Use GNUInstallDirs for install destination

### LINKED ISSUE #188: The `* .dll` files installed in `lib` directory (should be `bin`)
After installation on DLL-based platform, the `*.dll` part of shared libraries will be placed in `lib` subdirectory, but it should be the `bin` directory.

podsvirov:
Closes #188.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,12 +88,43 @@ cd build
 cmake -DGTEST_ROOT=/usr/src/gtest ..
 make
 make test  # If GTEST_ROOT specified above.
-sudo make install
 ```
 
 Enable gflags and glog with `cmake -DWITH_GFLAGS=ON -DWITH_GLOG=ON ...`.
 
 Disable building of shared libraries with `-DBUILD_SHARED_LIBS=OFF`.
+
+## Installing
+
+From `build` subdirectory:
+
+```
+make install
+```
+
+Prefix it with `sudo` if needed:
+
+```
+sudo make install
+```
+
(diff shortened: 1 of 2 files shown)


jmr:
Can you add something to README.md about how to configure install locations?

podsvirov:
> Can you add something to README.md about how to configure install locations?

Done. Please review again (sorry for possible  lingual mistakes).

README.md
```diff
@@ -88,13 +88,44 @@ cd build
...
+make install
+```
+
+Prefix it with `sudo` if needed:
+
+```
+sudo make install
+```
+
+_NOTE_: The `uninstall` target missed but `install_manifest.txt` may be helpfull.
```
jmr, line 111: How about "there is not `uninstall` target but ..."?

README.md
```diff
@@ -88,13 +88,44 @@ cd build
...
+
+Several suffix variables used for some file groups:
+
+Variable | Description
+-------- | -----------
+`CMAKE_INSTALL_INCLUDEDIR` | For header files
+`CMAKE_INSTALL_BINDIR`     | For executables and `*.dll` files on `DLL`-based platforms
+`CMAKE_INSTALL_LIBDIR`     | For library files (`*.so`, `*.a`, `*.lib` etc)
+
+To edit cached variables type:
```
jmr, line 123: What are cached variables? I'm not sure this is going to help without further explanation.

jmr (changes requested):

PR #189, opened 2021-04-30
Make python optional via flags

sammorris81:
Add a flag to determine if we should install the Python interface, and add additional python3-dev dependency to the install instructions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,8 +97,7 @@ include_directories(
     ${GFLAGS_INCLUDE_DIRS} ${GLOG_INCLUDE_DIRS} ${OPENSSL_INCLUDE_DIR})
 
 if (WITH_PYTHON)
-  include_directories(
-    ${Python3_INCLUDE_DIRS})
+    include_directories(${Python3_INCLUDE_DIRS})
 endif()
 
 include_directories(src)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -88,8 +94,13 @@ endif()
 # something like:
 # OPENSSL_ROOT_DIR=/usr/local/opt/openssl cmake ..
 include_directories(
-    ${GFLAGS_INCLUDE_DIRS} ${GLOG_INCLUDE_DIRS} ${OPENSSL_INCLUDE_DIR}
+    ${GFLAGS_INCLUDE_DIRS} ${GLOG_INCLUDE_DIRS} ${OPENSSL_INCLUDE_DIR})
+
+if (WITH_PYTHON)
+  include_directories(
```
jmr, line 100: put this on one line since it fits now

README.md
```diff
@@ -133,11 +133,12 @@ For more info read: [The CMake Cache](https://cmake.org/cmake/help/latest/guide/
 If you want the Python interface, you will also need:
```
jmr, line 133: you'll also need to run cmake with -DWITH_PYTHON

jmr (changes requested):

sammorris81:
All suggestions addressed. Please let me know if there's anything additional.

jmr:
Thanks for the improvement/fix!

jmr:
Sorry, one more thing I just thought of. Can you add python back here, so it's still tested, since it now defaults to off??

https://github.com/google/s2geometry/blob/master/.travis.yml#L162

sammorris81:
Added -DWITH_PYTHON=ON to the test. Please let me know if there's anything additional you would like.

PR #220, opened 2021-11-19
BN_num_bits takes a uint64_t*, so pass it one.

nlewycky:
Somehow, on my system `uint64` from port.h and `uint64_t` are typedefs of two different 64-bit types, `unsigned long` and `unsigned long long`.

For the most part this difference would be invisible, however when passing by pointer the pointer must point to the correct type.

No functionality change.


src/s2/util/math/exactfloat/exactfloat.cc
```diff
@@ -71,7 +71,7 @@ inline static void BN_ext_set_uint64(BIGNUM* bn, uint64 v) {
 // Return the absolute value of a BIGNUM as a 64-bit unsigned integer.
 // Requires that BIGNUM fits into 64 bits.
 inline static uint64 BN_ext_get_uint64(const BIGNUM* bn) {
-  uint64 u64;
+  std::uint64_t u64;
```
jmr, line 74: just use `uint64_t`. it should work everywhere.

nlewycky: Done.

Issue #237, opened 2022-03-29
use of undeclared identifier 'ABSL_INTERNAL_ASSUME'


t-brito:
As of [this commit](https://github.com/abseil/abseil-cpp/commit/231c393a170ce3c6d83e8456bc87fe917c333ecf), abseil no longer includes the `ABSL_INTERNAL_ASSUME` macro, which causes `s2geometry` build to fail on several places inside `src/s2/util/bits/bits.h` with the error message:

`error: use of undeclared identifier 'ABSL_INTERNAL_ASSUME'`

I will open a PR to fix this by replacing `ABSL_INTERNAL_ASSUME` with `ABSL_ASSUME`

PR #238, opened 2022-03-29
fixing error: use of undeclared identifier 'ABSL_INTERNAL_ASSUME'

t-brito:
Proposed fix for https://github.com/google/s2geometry/issues/237

### DIFF
--- a/src/s2/util/bits/bits.h
+++ b/src/s2/util/bits/bits.h
@@ -58,6 +58,10 @@
 #if defined(_MSC_VER)
 #include <intrin.h>
 #endif
+// keep compatibility with abseil-cpp LTS branch "lts_2021_11_02"
+#if !defined(ABSL_ASSUME)
+#define ABSL_ASSUME(cond) ABSL_INTERNAL_ASSUME(cond)
+#endif
 
 #include <type_traits>
 
@@ -362,7 +366,7 @@ inline int Bits::Log2Floor(T &&n) {
 }
 
 inline int Bits::Log2FloorNonZero(uint32 n) {
-  ABSL_INTERNAL_ASSUME(n != 0);
+  ABSL_ASSUME(n != 0);
   return absl::bit_width(n) - 1;
 }
 
@@ -384,17 +388,17 @@ inline int Bits::Log2Floor64(T &&n) {
 }
 
 inline int Bits::Log2FloorNonZero64(uint64 n) {
-  ABSL_INTERNAL_ASSUME(n != 0);
+  ABSL_ASSUME(n != 0);
   return absl::bit_width(n) - 1;
 }
 
 inline int Bits::FindLSBSetNonZero(uint32 n) {
-  ABSL_INTERNAL_ASSUME(n != 0);
+  ABSL_ASSUME(n != 0);
   return absl::countr_zero(n);
 }
 
 inline int Bits::FindLSBSetNonZero64(uint64 n) {
-  ABSL_INTERNAL_ASSUME(n != 0);
+  ABSL_ASSUME(n != 0);
   return absl::countr_zero(n);
 }
 
--- a/src/s2/util/bits/bits.h
+++ b/src/s2/util/bits/bits.h
@@ -58,9 +58,14 @@
 #if defined(_MSC_VER)
 #include <intrin.h>
 #endif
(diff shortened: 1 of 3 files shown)


jmr:
How long has `ABSL_ASSUME` existed? It would be good if this worked with the last LTS abseil-cpp release.

t-brito:
Not that long in fact.

This [commit](https://github.com/abseil/abseil-cpp/commit/5e4ea1ce097f3571e7d87af33b6b30d11b3a211e) from 24 February 2022 creates `ABSL_ASSUME` as a replacement to `ABSL_INTERNAL_ASSUME` with the message:

`Make ABSL_ASSUME publicly available`

Then on 16 March 2022 `ABSL_INTERNAL_ASSUME` gets removed completely in the [commit](https://github.com/abseil/abseil-cpp/commit/231c393a170ce3c6d83e8456bc87fe917c333ecf) mentioned above, with the message:

`Remove the internal-only ABSL_INTERNAL_ASSUME now that ABSL_ASSUME
is available and ABSL_INTERNAL_ASSUME has no more users.`

I'll add a definition which was present in Abseil's code during the transition

src/s2/util/bits/bits.h
```diff
@@ -58,6 +58,10 @@
 #if defined(_MSC_VER)
 #include <intrin.h>
 #endif
+// keep compatibility with abseil-cpp LTS branch "lts_2020_09_23"
+#if !defined(ABSL_ASSUME)
+#define ABSL_ASSUME(cond) ABSL_INTERNAL_ASSUME(cond)
```
t-brito, line 63: Using the same definition as [here](https://github.com/abseil/abseil-cpp/blame/5ed77665c4d697f657fa1362b5a482450f858cdc/absl/base/optimization.h#L219), which should provide the same behaviour for code older than the definition of `ABSL_ASSUME `

jmr: It makes me a bit uneasy to mess with `ABSL_` symbols in a header. How about something like

```
// TODO: Remove this and just use ABSL_ASSUME when it is in an LTS release.
#if defined(ABSL_ASSUME)
#define S2_ASSUME(cond) ABSL_ASSUME(cond)
#elif defined(ABSL_INTERNAL_ASSUME)
#define #S2_ASSUME ...
#else
#error "abseil-cpp must provide ABSL_ASSUME or ABSL_INTERNAL_ASSUME, what version are you using?"
#endif
```
?

t-brito: I understand your point, I was just taking a bit of code directly from the original project to make sure I copied the intended behaviour

To be clear, this suggestion would also involve replacing all calls to `ABSL_INTERNAL_ASSUME` with `S2_ASSUME` and eventually replace those with `ABSL_ASSUME ` right?

jmr: Yes, thanks!

jmr:
Did you test the final version (with any absl is fine)? Moving from travis-ci to github actions is still on my TODO...

t-brito:
Ran `make test` with both `abseil-cpp` [lts_2021_11_02](https://github.com/abseil/abseil-cpp/tree/lts_2021_11_02/absl) and `master` (current commit [4dcae40a07](https://github.com/abseil/abseil-cpp/commit/4dcae40a07e30cbc99d4130e590047000e5fda0b)...)

Make settings:

```
cmake \
 -DPython3_EXECUTABLE:FILEPATH=.../bin/python3.7 \
 -DPYTHON_EXECUTABLE=.../bin/python3.7 \
 -DWITH_GFLAGS=ON \
 -DWITH_GLOG=ON \
 -DBUILD_SHARED_LIBS=OFF \
 -DWITH_PYTHON=ON \
 -WITH_GTEST=ON \
 -DGTEST_ROOT=.../googletest-release-1.8.0/googletest \
 -DCMAKE_PREFIX_PATH=abseil_[master|lts] \
 -DOPENSSL_ROOT_DIR=.../openssl@3 \
 -DOPENSSL_INCLUDE_DIR=.../openssl@3/include \
 -DCMAKE_CXX_STANDARD=11 \
 ..

```

All 107 tests passed with both `abseil-cpp` versions, some warnings related to C++11 I believe

jmr:
Great, thanks for the fix!

PR #249, opened 2022-05-04
Support building in -std=c++20 mode.

pkasting:
Types on both sides of comparison operators must be the same.

Bug: chromium:1284275


jmr:
What's the error you get without this PR?

jmr:
#254 should fix this.

PR #253, opened 2022-05-17
Missing include statement for abseil Hash

marcus-elia:
I ran into this error
`src/s2/s2cell_union.h:413:41: error: no member named 'HashOf' in namespace 'absl'`
This PR adds
`#include "absl/hash/hash.h`


jmr:
#254 adds this.

PR #257, opened 2022-05-26
Expose more functionality via the SWIG wrapper

MikePlayle:
Add S2::UpdateMinDistance, s2pred::OrderedCCW, S2::Interpolate,
S2Polygon::InitToUnion, some more of S2::S1ChordAngle, and a
method for retrieving the S2Point's underlying 3-vector.

### DIFF
--- a/src/python/pywraps2_test.py
+++ b/src/python/pywraps2_test.py
@@ -787,6 +787,13 @@ class PyWrapS2TestCase(unittest.TestCase):
     radius_m = s2.S2Earth.RadiansToMeters(angle.radians())
     self.assertEqual(radius_m, 12340.0)
 
+  def testS2Point_ToFromRaw(self):
+    p = s2.S2Point_FromRaw(1.0, 1.0, 1.0)
+    p = p.Normalize()
+    p_raw = s2.S2Point_ToRaw(p)
+    self.assertAlmostEqual(p_raw[0], 0.57735027)
+    self.assertAlmostEqual(p_raw[1], 0.57735027)
+    self.assertAlmostEqual(p_raw[2], 0.57735027)
 
 class RegionTermIndexerTest(unittest.TestCase):
   def _randomCaps(self, query_type, **indexer_options):
--- a/src/python/pywraps2_test.py
+++ b/src/python/pywraps2_test.py
@@ -1158,5 +1158,66 @@ class S2BooleanOperationTest(unittest.TestCase):
     loop = result.loop(1)
     self.assertEqual(4, loop.num_vertices())
 
+class S2BuilderTest(unittest.TestCase):
+  def setUp(self):
+    self.p1 = s2.S2LatLng.FromDegrees(10.0, 10.0).ToPoint()
+    self.p2 = s2.S2LatLng.FromDegrees(10.0, 11.0).ToPoint()
+    self.p3 = s2.S2LatLng.FromDegrees(11.0, 10.0).ToPoint()
+    self.p4 = s2.S2LatLng.FromDegrees(11.0, 11.0).ToPoint()
+
+  def testSquareDirected(self):
+    opts = s2.S2PolygonLayerOptions()
+    opts.set_edge_type(s2.S2Builder.EdgeType_DIRECTED)
+
+    result = s2.S2Polygon()
+    b = s2.S2Builder()
+    b.StartLayer(s2.S2PolygonLayer(result, opts))
+
+    b.AddEdge(self.p1, self.p2)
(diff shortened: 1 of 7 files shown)


MikePlayle:
Extended to add S2BufferOperation, S2BooleanOperation and S2Builder, along with various things that they depend on.

I'm not sure that this is necessarily the "right way" to do it - if there are any problems here, now would be a good time to discuss them.

MikePlayle:
Is there anything else I need to do to this pull request?

Would it help if I split it up?

jmr:
Splitting it up would probably get you pieces sooner, if that's useful to you. Probably not worth the effort otherwise. I'll look more at this tomorrow.

src/python/s2_common.i
```diff
@@ -372,6 +498,9 @@ class S2Point {
 
 %ignoreall
 
+%unignore MutableS2ShapeIndex;
+%unignore MutableS2ShapeIndex::~MutableS2ShapeIndex;
+%unignore MutableS2ShapeIndex::Add(S2Shape*);
```
jmr, line 503: There are a lot of functions added. Are they all tested? There are too many to easily tell.

src/python/pywraps2_test.py
```diff
@@ -951,5 +951,185 @@ def testS2CellIdFromDebugString(self):
     self.assertTrue(cell.is_valid())
     self.assertEqual("5/31200", cell.ToString())
 
+class ChordAngleTest(unittest.TestCase):
```
jmr, line 954: Having test classes is consistent with the rest of the file, but I like it. This would normally be called `S2ChordAngleTest`

src/python/s2_common.i
```diff
@@ -19,10 +25,36 @@
...
 #include "s2/s2polyline.h"
+#include "s2/s2predicates.h"
 #include "s2/s2region_coverer.h"
 #include "s2/s2region_term_indexer.h"
 #include "s2/s2cell.h"
 #include "s2/s2cell_union.h"
+#include "s2/s2shape_index.h"
+#include "s2/mutable_s2shape_index.h"
+
+class S2BufferOperationOptions {
```
jmr, line 36: I assume this is a work-around for nested classes. Add comments for why this stuff is here.

jmr (changes requested):
I've only just started to look at this. I'll have more questions later, since not everything is clear to me.

src/python/s2_common.i
MikePlayle: Yes - see the commit message. I'll add a similar comment to S2PolygonLayerOptions too.

src/python/pywraps2_test.py
MikePlayle: This is normal in Python unittest code, in my experience at least.

I chose the names following the example of `RegionTermIndexerTest`, but I'm happy to change them.

src/python/s2_common.i
MikePlayle: I hope so, but I'll review the changes to make sure. (I'll leave this open for now to remind me to do it.)

MikePlayle: Oh, I realised that I didn't add tests for the ones that were basically just a `%unignore` - is this what you meant? I can add these if you like.

MikePlayle: I think everything is covered now.

src/python/s2_common.i
```diff
@@ -71,6 +107,15 @@
 
 %apply S2CellId *OUTPUT_ARRAY_4 {S2CellId neighbors[4]};
 
+%typemap(in, numinputs=0) S2Error * (S2Error err) {
```
jmr, line 110: What does this do / where is it used? Maybe add a comment since it's not obvious.

jmr:
Sorry for the delay, I'll merge after I understand https://github.com/google/s2geometry/pull/257/files#r953600009 .

src/python/s2_common.i
MikePlayle: This pair of typemaps work together to translate errors from the C++ library into Python exceptions. The first declares a local variable `S2Error err;` and sets the matched argument (`$1`, of type `S2Error *`) to point to it; the second inspects the value of the `S2Error` after the call and throws an exception if its `ok()` method returns false. `S2BuilderTest.testException` in `pywraps2_test.py` tests this.

https://www.swig.org/Doc1.3/Typemaps.html#Typemaps_nn23 describes the syntax being used here.

I've added a comment.

jmr:
Thanks!

PR #267, opened 2022-08-11
Fix RemoveLast

MBkkt:


jmr:
Can you explain how it's broken?

MBkkt:
unary minus operator applied to unsigned type, result still unsigned

I don't think it's desirable

jmr:
It's not broken unless `ptrdiff_t` and `size_t` aren't the same size, but I agree it's clearer your way.

PR #268, opened 2022-08-20
python: Named component accessors for S2Point

podsvirov:
With this small change we can code something like this:

```python
import numpy as np
from pywraps2 import S2Cell, S2CellId
from matplotlib import pyplot as plt

def surf4face(f, l):
    m = 2**l; m1 = m + 1; m2 = m - 1
    k = range(0, 2**S2CellId.kMaxLevel, 2**(S2CellId.kMaxLevel - l))
    x = np.empty((m1,m1)); y = np.empty((m1,m1)); z = np.empty((m1,m1));
    for i in range(m):
        for j in range(m):
            c = S2Cell(S2CellId.FromFaceIJ(f, k[i], k[j]).parent(l))
            v = c.GetVertex(0)
            x[i][j] = v.x(); y[i][j] = v.y(); z[i][j] = v.z();
            if i == m2:
                v = c.GetVertex(1)
                x[m][j] = v.x(); y[m][j] = v.y(); z[m][j] = v.z();
            if j == m2:
                v = c.GetVertex(3)
                x[i][m] = v.x(); y[i][m] = v.y(); z[i][m] = v.z();
            if i == m2 and j == m2:
                v = c.GetVertex(2)
                x[m][m] = v.x(); y[m][m] = v.y(); z[m][m] = v.z();
    return (x, y, z)

fig = plt.figure()
ax = fig.add_subplot(projection='3d')

for f in range(6):
    s = surf4face(f, 2)
    ax.plot_surface(*s, color='blue')
    ax.plot_wireframe(*s, color='green')

fig.savefig('s2sphere.svg')
```

to render this:

s2sphere

and not only...

### DIFF
--- a/src/python/pywraps2_test.py
+++ b/src/python/pywraps2_test.py
@@ -22,6 +22,15 @@ import pywraps2 as s2
 
 class PyWrapS2TestCase(unittest.TestCase):
 
+  def testS2PointFromRawToNamedCorrectly(self):
+      x = 1.0
+      y = 2.0
+      z = 3.0
+      point = s2.S2Point_FromRaw(x, y, z)
+      self.assertEqual(x, point.x())
+      self.assertEqual(y, point.y())
+      self.assertEqual(z, point.z())
+
   def testContainsIsWrappedCorrectly(self):
     london = s2.S2LatLngRect(s2.S2LatLng.FromDegrees(51.3368602, 0.4931979),
                              s2.S2LatLng.FromDegrees(51.7323965, 0.1495211))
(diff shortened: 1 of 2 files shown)


MikePlayle:
This overlaps with a change that I proposed: https://github.com/google/s2geometry/pull/257/commits/e8f4ae4cc7b5fb8343cd9e684db78f7bd54579cb

I don't know which way is better. I'd be happy with either.

podsvirov:
The changes do not conflict, but complement each other. I hope that one small change will be quickly reviewed and accepted.

jmr:
Please add a test to [`pywraps2_test.py`](https://github.com/google/s2geometry/blob/master/src/python/pywraps2_test.py) so these new functions don't get broken in the future.

podsvirov:
> Please add a test to [`pywraps2_test.py`](https://github.com/google/s2geometry/blob/master/src/python/pywraps2_test.py) so these new functions don't get broken in the future.

Done. Please review again.

PR #273, opened 2022-09-13
Fix -DCMAKE_CXX_STANDARD ignored by CMakeLists

### LINKED ISSUE #270: -DCMAKE_CXX_STANDARD=17 doesn't work
The cache of build is cleared, and I specify `-DCMAKE_CXX_STANDARD=17`, but s2 still use c++11 to compile.
But if I change the line `set(CMAKE_CXX_STANDARD 11)` to `set(CMAKE_CXX_STANDARD 17)` in CmakeLists.txt, it works.

LafeWessel:
The command line argument for setting the C++ standard to use is ignored. This change allows the default to remain 11 but allow for the user to override it.

Resolve #270

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,8 +18,9 @@ endif()
 # s2geometry needs to use the same C++ standard that absl used to avoid
 # undefined symbol errors since ABSL_HAVE_STD_STRING_VIEW etc will
 # end up defined differently.  There is probably a better way to achieve
-# this than assuming what absl used.
-set(CMAKE_CXX_STANDARD 11)
+# this than assuming what absl used. 
+# Using CACHE allows the user to override the default.
+(CMAKE_CXX_STANDARD 11 CACHE)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 # No compiler-specific extensions, i.e. -std=c++11, not -std=gnu++11.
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -18,8 +18,9 @@ endif()
 # s2geometry needs to use the same C++ standard that absl used to avoid
 # undefined symbol errors since ABSL_HAVE_STD_STRING_VIEW etc will
 # end up defined differently.  There is probably a better way to achieve
-# this than assuming what absl used.
-set(CMAKE_CXX_STANDARD 11)
+# this than assuming what absl used. 
+# Using CACHE allows the user to override the default.
+set(CMAKE_CXX_STANDARD 11 CACHE STRING "The C++ standard to build with")
```
jmr, line 23: https://github.com/google/s2geometry/issues/270#issuecomment-1235346863 suggests `option`. What's the difference between `set(CACHE)` and `option` / when would I want to use one or the other?  @jievince

jievince: Hi @jmr, I am unfamiliar with CMake, and I don't know the difference between them. I just tried `option`, and it works.

jievince: Hi @LafeWessel, can you explain it?

jievince (approved):
LGTM

Issue #278, opened 2022-10-25
ABI compatability and SOVERSION


spaetz:
Hello,
I am about to finish the Debian packaging of s2geometry, so the lib will likely make it into the Debian repository soon. One  question on your ABI stability promises:

CMakeLists.txt contains `SOVERSION ${PROJECT_VERSION_MAJOR}`, which a Debian developer has called "optimistic". Are you planning to remain ABI compatible over all of `${PROJECT_VERSION_MAJOR}`? If yes, great, if no, we can still change that :)

Thanks for s2geometry

spaetz:
P.S. Just FYI: s2geometry is about to find its way into main Debian right now:
https://ftp-master.debian.org/new/s2geometry_0.10.0-2.html

jmr:
> Are you planning to remain ABI compatible over all of ${PROJECT_VERSION_MAJOR}?

I was planning to, once it reached version 1, but we haven't promised anything. Definitely no promises for version 0.

> CMakeLists.txt contains SOVERSION ${PROJECT_VERSION_MAJOR}, which a Debian developer has called "optimistic".

Yeah, this would probably mean that nearly every release is a new major version.

Do you have any suggestions of best practices? Should we remove `SOVERSION`?

spaetz:
Right, now you need to bump the major version of all ABI breaks. You can just do that and keep things that way, but
how about simply manually setting
SOVERSION 1
and bumping it manually when you break the ABI? Or removing it, and not promising anything at all....

Both options are possible, I guess.

jmr:
> how about simply manually setting SOVERSION 0 and bumping it manually when you break the ABI?

We'll update the project version appropriately.

> Or remove SOVERSION

Looks like `SOVERSION` was added in #161, so some people might not like that.

According to [SemVer, version 0 promises nothing](https://semver.org/#spec-item-4). Is that ok for you?

spaetz:
Yes, that would be fine with us.

jmr:
In addition to the general "version 0 promises nothing" warning, I'd like to call out that we want to change the python bindings significantly (like every function name changes) soon.  

https://ftp-master.debian.org/new/s2geometry_0.10.0-2.html isn't there anymore, but it may be better to hold off on packaging the python parts so we don't get a bunch of new users who are then soon broken.  @spaetz What do you think?

PR #281, opened 2022-11-17
README.md: Add warning about API/ABI stability

jmr:
#278 (ABI compatability and SOVERSION) asked questions about ABI stability.

Clarify that nothing is stable while we are on 0.x releases, but we will adhere to SemVer starting with 1.0.


jmr:
@eengle @spaetz

spaetz:
Good with me, thanks

PR #284, opened 2022-11-22
Fix undefined sanitizer issue in compact_array

MBkkt:
Also I think really correct code, should start with non zero capacity (so correct capacity can be seted in ctor)

### DIFF
--- a/src/s2/util/gtl/compact_array.h
+++ b/src/s2/util/gtl/compact_array.h
@@ -415,8 +415,8 @@ class compact_array_base {
     T* new_ptr = allocator.allocate(capacity());
     if (old_capacity != 0) {
       memcpy(new_ptr, Array(), old_capacity * sizeof(T));
+      allocator.deallocate(Array(), old_capacity);
     }
-    allocator.deallocate(Array(), old_capacity);
 
     SetArray(new_ptr);
   }
--- a/src/s2/util/gtl/compact_array.h
+++ b/src/s2/util/gtl/compact_array.h
@@ -413,7 +413,9 @@ class compact_array_base {
     value_allocator_type allocator;
 
     T* new_ptr = allocator.allocate(capacity());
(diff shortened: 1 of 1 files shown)


src/s2/util/gtl/compact_array.h
```diff
@@ -413,7 +413,9 @@ class compact_array_base {
     value_allocator_type allocator;
 
     T* new_ptr = allocator.allocate(capacity());
-    memcpy(new_ptr, Array(), old_capacity * sizeof(T));
+    if (old_capacity != 0) {
+      memcpy(new_ptr, Array(), old_capacity * sizeof(T));
+    }
     allocator.deallocate(Array(), old_capacity);
```
MBkkt, line 419: Maybe better? I don't read class code

```suggestion
    if (old_capacity != 0) {
      memcpy(new_ptr, Array(), old_capacity * sizeof(T));
      allocator.deallocate(Array(), old_capacity);
    }
```

jmr:
> Also I think really correct code, should start with non zero capacity (so correct capacity can be seted in ctor)

Why? Feel free to send another PR if you think there are more improvements.

MBkkt:
> Why? Feel free to send another PR if you think there are more improvements

Because we always have some capacity on "stack". Same as for sso string.

jmr:
> Because we always have some capacity on "stack". Same as for sso string.

I'm not sure what you mean. There's already inline SSO-like space allocated: https://github.com/google/s2geometry/blob/e15d00b21e9bbc252e436c243710cb9712465814/src/s2/util/gtl/compact_array.h#L124

Send a PR to make it clear.

MBkkt:
I mean sso string after default ctor have non zero capacity.
Why same not possible for this array?

jmr:
Anything is possible. Send a PR if you think it can be improved.

PR #285, opened 2022-11-28
Allow already have abseil

MBkkt:
We don't install abseil, we use add_subdirectory. so for us I needs patch like this.

Is it ok for you? The main idea if we already have abseil targets we don't need to find them

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -68,7 +68,9 @@ if (WITH_GFLAGS)
     add_definitions(-DS2_USE_GFLAGS)
 endif()
 
-find_package(absl REQUIRED)
+if (NOT TARGET absl::base)
+    find_package(absl REQUIRED)
+endif()
 find_package(OpenSSL REQUIRED)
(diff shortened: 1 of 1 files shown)


smcallis:
This seems minor enough to be fine.  Abseil should probably support this workflow upstream too.  Wish CMake made this easier.

MBkkt:
Thanks!

CMakeLists.txt
```diff
@@ -68,7 +68,9 @@ if (WITH_GFLAGS)
     add_definitions(-DS2_USE_GFLAGS)
 endif()
 
-find_package(absl REQUIRED)
+if (NOT TARGET absl::base)
```
jmr, line 71: If I just saw this code without having seen the PR, I would have no idea what it's doing or why.

Could you add a comment?

MBkkt: I will add

MBkkt: https://github.com/google/s2geometry/pull/286

PR #287, opened 2022-12-14
Fix possible warning in header

MBkkt:
if no defined ABSL_MIN_LOG_LEVEL severity_ will be unused.

Because its header it's pretty inconvenient.

As an alternative solution will be make field under ifdef.
But in this case layout of class will be changed, so I don't know, is it valid for you, I decide to make more simple fix.

### DIFF
--- a/src/s2/base/logging.h
+++ b/src/s2/base/logging.h
@@ -62,7 +62,7 @@ class S2LogMessage {
     : severity_(severity), stream_(stream) {
     if (enabled()) {
       stream_ << file << ":" << line << " "
-              << absl::LogSeverityName(severity) << " ";
+              << absl::LogSeverityName(severity_) << " ";
     }
   }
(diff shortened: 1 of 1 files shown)


smcallis:
Seems like a reasonable workaround.

jmr:
What does "possible warning" mean? Is there some compiler/set of options you actually observed a warning with?

This will all be deleted soon anyway, since absl now supports logging.

MBkkt:
> What does "possible warning" mean? Is there some compiler/set of options you actually observed a warning with?

Yes, gcc 11.2 with some flags, warnings about unused variable. But I think it can be avoided if we interprete s2 headers like system headers.

> This will all be deleted soon anyway, since absl now supports logging.

Good we don't really use it.

MBkkt:
https://github.com/arangodb/arangodb/blob/devel/3rdParty/README_maintainers.md#s2geometry

We have patch, and mostly it fix warnings

PR #291, opened 2022-12-19
Fix missing include

MBkkt:
https://github.com/google/s2geometry/blob/master/src/s2/s2polygon.cc#L1293


jmr:
I'm not sure why this is using a `multi_map` instead of a `priority_queue`, but looks good for now. Thanks.

jmr:
That would make sense, but this was written before `unique_ptr` was used.

PR #298, opened 2023-01-09
Fix address sanitizer issue in S2Loop move ctor/assign

MBkkt:
Code copied from S2Polygon

### DIFF
--- a/src/s2/s2loop.cc
+++ b/src/s2/s2loop.cc
@@ -110,7 +110,14 @@ S2Loop::S2Loop(S2Loop&& b)
           b.unindexed_contains_calls_.exchange(0, std::memory_order_relaxed)),
       bound_(std::move(b.bound_)),
       subregion_bound_(std::move(b.subregion_bound_)),
-      index_(std::move(b.index_)) {}
+      index_(std::move(b.index_)) {
+  // `index_` has a pointer to an S2Loop::Shape which points to S2Loop.
+  // But, we've moved to a new address, so get the Shape back out of the index
+  // and update it to point to our new location.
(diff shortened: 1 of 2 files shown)


src/s2/s2loop.cc
```diff
@@ -110,7 +110,14 @@ S2Loop::S2Loop(S2Loop&& b)
           b.unindexed_contains_calls_.exchange(0, std::memory_order_relaxed)),
       bound_(std::move(b.bound_)),
       subregion_bound_(std::move(b.subregion_bound_)),
-      index_(std::move(b.index_)) {}
+      index_(std::move(b.index_)) {
+  // `index_` has a pointer to an S2Loop::Shape which points to S2Loop.
+  // But, we've moved to a new address, so get the Shape back out of the index
+  // and update it to point to our new location.
+  if (index_.begin() != index_.end()) {
```
smcallis, line 117: You can just do:

```
for (S2Shape* shape : index_) { ...} 
```

rather than checking the iterators directly.

smcallis (commented):
Nice catch, I'll fix this internally too.

src/s2/s2loop.cc
MBkkt: Yes, but I just copied it from S2Polygon.

I think for your suggestion better to separate PR with same change in S2Polygon

smcallis: Yup you're probably right =D

MBkkt:
@jmr Is it ok?

jmr:
Yes, it will be merged after #297

PR #301, opened 2023-01-09
Add S2_DO_NOT_INSTALL option with keeping old behavior

MBkkt:


jmr:
How is this different from just never typing `make install`?

MBkkt:
Not really I think?

Ok, closed

PR #303, opened 2023-01-09
Term indexer improvements

MBkkt:


src/s2/s2region_term_indexer_test.cc
```diff
@@ -84,7 +84,7 @@ void TestRandomCaps(const S2RegionTermIndexer::Options& options,
     // random size.
     S2Cap cap;
     vector<string> terms;
-    if (query_type == QueryType::CAP) {
+    if (query_type == QueryType::POINT) {
```
MBkkt, line 87: Test was correct but when was written
```
TestRandomCaps(options, QueryType::CAP);
```
it meant query points

and otherwise
```
TestRandomCaps(options, QueryType::Point);
```
meant caps

I think it's typo

src/s2/s2region_term_indexer.cc
```diff
@@ -194,29 +215,34 @@ vector<string> S2RegionTermIndexer::GetIndexTermsForCanonicalCovering(
...
 vector<string> S2RegionTermIndexer::GetQueryTerms(const S2Point& point,
                                                   string_view prefix) {
+  vector<string> terms;
+  GetQueryTerms(point, prefix, &terms);
+  return terms;
+}
+
+void S2RegionTermIndexer::GetQueryTerms(const S2Point& point,
+                                        string_view prefix,
+                                        vector<string>* terms) {
```
jmr, line 233: Please split changing the type signature into a separate PR. We can discuss it there. Does this have a measurable and meaningful performance difference for you? I would have guessed that computing the covering takes way longer than reallocating vectors.

https://google.github.io/eng-practices/review/developer/small-cls.html

src/s2/s2region_term_indexer.cc
```diff
@@ -136,12 +144,13 @@ vector<string> S2RegionTermIndexer::GetIndexTerms(const S2Point& point,
   // max_level() != true_max_level() (see S2RegionCoverer::Options).
 
   const S2CellId id(point);
-  vector<string> terms;
-  for (int level = options_.min_level(); level <= options_.max_level();
-       level += options_.level_mod()) {
-    terms.push_back(GetTerm(TermType::ANCESTOR, id.parent(level), prefix));
+  int level = options_.min_level();
+  if (options_.query_contains_points_only()) {
```
jmr, line 148: Update "Optimizations" section of indexing strategy above.

src/s2/s2region_term_indexer.cc
```diff
@@ -245,18 +278,19 @@ vector<string> S2RegionTermIndexer::GetQueryTermsForCanonicalCovering(
     S2_DCHECK_GE(level, options_.min_level());
     S2_DCHECK_LE(level, options_.max_level());
     S2_DCHECK_EQ(0, (level - options_.min_level()) % options_.level_mod());
+    // assume level <= options_.true_max_level()
```
jmr, line 281: Change comment to `S2_DCHECK_LE`.

src/s2/s2region_term_indexer.cc
```diff
@@ -245,18 +278,19 @@ vector<string> S2RegionTermIndexer::GetQueryTermsForCanonicalCovering(
...
 
     // If the index only contains points, there are no covering terms.
     if (options_.index_contains_points_only()) continue;
 
     // If we are optimizing for index space rather than query time, cells are
     // also queried as covering terms (except for true_max_level() cells,
     // which are indexed and queried as ancestor cells only).
-    if (options_.optimize_for_space() && level < true_max_level) {
-      terms.push_back(GetTerm(TermType::COVERING, id, prefix));
+    if (options_.optimize_for_space() && level != true_max_level) {
```
jmr, line 292: Only change what you need to. More changes just makes it harder to see what's really changing.

src/s2/s2region_term_indexer_test.cc
```diff
@@ -122,6 +122,9 @@ TEST(S2RegionTermIndexer, IndexRegionsQueryRegionsOptimizeTime) {
   options.set_max_level(16);
   options.set_max_cells(20);
   TestRandomCaps(options, QueryType::CAP);
+  TestRandomCaps(options, QueryType::POINT);
```
jmr, line 125: This test case is called `IndexRegionsQueryRegions...`, so don't query points here.

src/s2/s2region_term_indexer_test.cc
jmr: Yes, it looks like it was wrong. Thanks for fixing it.

src/s2/s2region_term_indexer_test.cc
```diff
@@ -131,6 +134,8 @@ TEST(S2RegionTermIndexer, IndexRegionsQueryPointsOptimizeTime) {
   options.set_max_level(16);
   options.set_max_cells(20);
   TestRandomCaps(options, QueryType::POINT);
+  options.set_query_contains_points_only(true);
```
jmr, line 137: I think we can remove `QueryType` and do it based only on `query_contains_points_only`.  This is similar to how it works for `index_contains_points_only`.

@smcallis what do you think?

jmr (changes requested):

src/s2/s2region_term_indexer_test.cc
MBkkt: > I think we can remove QueryType and do it based only on query_contains_points_only. This is similar to how it works for index_contains_points_only.

No. We want to test which query points with and without points only optimization

src/s2/s2region_term_indexer_test.cc
MBkkt: Ok, maybe needs to rename tests.

In general I think will be a lot of better to always test 3 cases:
query regions
query points
query points with points only optimization

src/s2/s2region_term_indexer.cc
MBkkt: I just want to make it same for different function

src/s2/s2region_term_indexer.cc
MBkkt: I don't want it, it's conflicting changes. It's not convenient to make separate PRs and then resolve conflict in one to another

src/s2/s2region_term_indexer.cc
MBkkt: Ok

src/s2/s2region_term_indexer.cc
MBkkt: Ok

src/s2/s2region_term_indexer.cc
jmr: Sure, it's better, but is it 10% better or 0.001% better? We need to weigh the complexity of the change vs. the performance improvements.  How much does this improve 1. time to build the whole index (or insert an object) 2. query time? (or some other user-observable action)

src/s2/s2region_term_indexer_test.cc
jmr: Why? Points are also regions, and this property is already used here: https://github.com/google/s2geometry/pull/303/files/0f9ef7a0ebe8d52ebf73cfb4c441cc1e3d553a05#diff-c9f79443ff2537a812dd6c149afd00960afcc88655f0514de8a1cc4741e78e79L66

I agree it would be more test coverage, but I'm fine making this same assumption we're already making.

MBkkt: In general exist such scenarios matrix:

index contains only points, only points with enabled optimization, anything
x
same about query 
x
time/space bool option


So 3*3*2 = 18 configuration

Do you agree?

src/s2/s2region_term_indexer.cc
MBkkt: > We need to weigh the complexity of the change vs. the performance improvements. 

But it's not affect users who doesn't want to use it

Why is it issue?
In general same api exists for S2RegionCoverer, simple and advanced

Why don't you want same for S2RegionCoverer?
https://github.com/google/s2geometry/blob/master/src/s2/s2region_coverer.h#L189

Could you explain please?

Why for coverer it's ok, but for indexer it's too difficult? They usage are pretty same

> Sure, it's better, but is it 10% better or 0.001% better? 

I don't measure this part, but it depends on what we do.

src/s2/s2region_term_indexer_test.cc
MBkkt: Ok I understood what do you mean, you are incorrect.

Now in master tested such cases:

index contains points: true/false
query points/query regions

index contains points optimization affect query terms!

So for opposite options I need to make same query (points only) with two different index

MBkkt: Anyway I wrote better coverage with automatic case combinations

src/s2/s2region_term_indexer.cc
MBkkt: Done

src/s2/s2region_term_indexer_test.cc
MBkkt: `16 (2 * 2 * 2 * 2) - 4 (2 * 2 * 1 * 1) = 12`
all combinations - skipped
as expected

src/s2/s2region_term_indexer_test.cc
MBkkt: Fixed

src/s2/s2region_term_indexer_test.cc
MBkkt: Hmm you right about now in my code wasn't checked query points but index contains caps.

Fixed.

`2 * 2 * 2 * 2 * 2 - 2 * (1 * 2 * 2 * 1 * 2) = 16`

cases now

src/s2/s2region_term_indexer.cc
jmr: > Why don't you want same as for S2RegionCoverer?

The reason there are two here is more historical than technical. The `vector*` version was there since ~2005, and the version returning `S2CellUnion` was added in 2017. In that time, the style guide changed from recommending pointer args for output parameters to recommending return values.

https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs

The reason it changed is that the performance difference is usually not big .

> I don't measure this part, but it depends on what we do.

Making performance optimizations without measurement is likely to misdirect effort.

If this were written from scratch, there would probably be no `vector*` version, or at least not until benchmarks demonstrated the utility.

MBkkt: I really don't understand what the problem I want to make single allocation for N point, now it's impossible, after this commit possible, it's extension.

It's not difficult or something...
It's same as other classes api (if it's only historical reason deprecate it! I don't think so, vector is more optimal)

MBkkt: https://github.com/google/s2geometry/pull/304 Done (I don't want to merge this branch, just push it for add ability to reproduce)

MBkkt: Same but maybe a little lower for covering if we already has it.

(Thats reason why I add new overload only for FromCanonical, because compute covering is expensive, and in that function it's unnecessary)

MBkkt: If I be honest I think it's possible to make benchmark better for old api. (in such case this benchmark became useless, because it will be reuse same memory (one in vector, another in allocator) in both overloads).

For points is possible to call exact vector::reserve, but I don't think it's good change because for regions it's impossible and it also makes new overload a little worse (unnecessary reserve)

And new overload make possible tuning on user side if necessary (if it's not necessary just use old)

jmr: > Measure any changes is impossible commonly we measure changes about what we cannot proof logical.

I'm not sure what you mean here. Why is measurement impossible?

> For an example using hashmap instead of linear search for more than few elements
> I don't think for such changes needs benchmarks

hashmap vs linear search are generally the same code complexity. `vector*` or `vector&` is more complex because now you have to document what happens to the existing state, have to clear it somewhere, etc.

> Has s2 any microbenchmarks? If it's too critical for you I can add another one for show the difference between this overloads

Yes, but not for these functions.

Microbenchmarks are a means, not an end. If you profiled your whole system, and found a meaningful amount of time spent in `S2RegionTermIndexer::Get*Terms`, then microbenchmarks could be used to direct optimizations.

MBkkt: > Why is measurement impossible?

measurement **all** impossible.
Because it costs time, and I have other tasks with more priority

MBkkt: > In that time, the style guide changed from recommending pointer args for output parameters to recommending return values

But it's not about style, it's different behavior.

I think (maybe I'm not right, so please correct me) style guide more about
`absl::Status DoSomething(...);` vs `bool/void DoSomething(..., absl::Status*);`

jmr: No, this case is covered, too. https://abseil.io/tips/176

Do the simpler thing unless there is *meaningful* performance difference.

MBkkt: Firstly it's tips not rules. Also it's about simple and advanced overloads, it's not like only one. So if developer have questions about it it can just don't use it, and use it only if really needed.
Secondly it's meaningful for me.

At least for points it's easy measurable change.
Of course you can suggest reserve, but in such case needs idea how to force allocator to don't reuse same memory in benchmark (because it won't be on practice)

jmr: > Secondly it's meaningful for me.

Ok, what are the numbers? How many terms do you have in average/worst case? How much time is it saving you? What is the larger operation you're trying to optimize and how long does that take?

MBkkt: Depends on user. For one case nothing another more

MBkkt: https://github.com/google/s2geometry/pull/303#discussion_r1066830083
https://github.com/google/s2geometry/pull/303#discussion_r1066833245

Waiting answer on other part of messages

MBkkt: What exactly numbers do I need to provide to convince you?
It's look not constructive now

You ignore all arguments about it's not only stylistic difference.

Or is it impossible?
And I need to maintain patch/fork?

jmr: > [#303 (comment)](https://github.com/google/s2geometry/pull/303#discussion_r1066830083) [#303 (comment)](https://github.com/google/s2geometry/pull/303#discussion_r1066833245)
> 
> Waiting answer on other part of messages

I don't see any questions there. What questions are you waiting for answers to?

> Depends on user. For one case nothing another more

Well, what's the range?

MBkkt: > Well, what's the range?

from 0 to thousands of inserts in each thread.
You see in benchmark difference like 20-30%, of course it's not all job, but I think even 1% for just call other API is good enough.

> I don't see any questions there. What questions are you waiting for answers to?

You sent a tip, in this tip have two caveats exactly about this situation, and also it have example with pretty same issue.

So I waiting why do you think this tips is  good for your arguments not for mine.

jmr: > You ignore all arguments about it's not only stylistic difference.

No, I agree it's more than style. We need to trade style against performance, so explain the performance impact.

> What exactly numbers do I need to provide to convince you?

Most convincing would be an arangodb load test where this change decreases latency by X%/T ms.

But you can start with how many index/query terms you're dealing with. Maybe the covering options are producing too many terms, and there's some other problem that should be solved.

MBkkt: > We need to trade style against performance, so explain the performance impact.

It's not trade. Old overloads still exist!

MBkkt: > But you can start with how many index/query terms you're dealing with. Maybe the covering options are producing too many terms, and there's some other problem that should be solved.

For an example I wrote count of inserts.
We have points only index, which can be created with default indexer options.

So one of example is exactly like in benchmark.

MBkkt: > It's not trade.

It's extension

MBkkt: https://github.com/abseil/abseil-cpp/blob/master/absl/strings/str_format.h

For an example it has two versions:
void StrAppendFormat and string StrFormat

jmr: > We have points only index, which can be created with default indexer options.

This part of the change is good.

> It's extension

You want to make a performance change. Explain the performance impact in a time unit.

MBkkt: > You want to make a performance change. Explain the performance impact in a time unit.

Already done

MBkkt: > This part of the change is good

I mean we have **index** with points only. Every allocation cost at least 10ns (on practice more), so if we make 1 instead of N it decrease overall time at least on N*10ns

Query with points only it's other part of PR.

src/s2/s2region_term_indexer_test.cc
```diff
@@ -112,62 +116,80 @@ void TestRandomCaps(const S2RegionTermIndexer::Options& options,
...
   options.set_max_cells(8);
-  TestRandomCaps(options, QueryType::CAP);
+  TestRandomCaps(options, index_type, query_type);
 }
 
-TEST(S2RegionTermIndexer, IndexPointsQueryRegionsOptimizeTime) {
-  S2RegionTermIndexer::Options options;
-  options.set_optimize_for_space(false);       // Optimize for time.
-  options.set_min_level(0);                    // Use face cells.
+TEST_P(S2RegionTermIndexerTest, UseFaceCells2) {
```
jmr, line 187: Use something more descriptive than "2" to differentiate it.

src/s2/s2region_term_indexer_test.cc
```diff
@@ -112,62 +116,80 @@ void TestRandomCaps(const S2RegionTermIndexer::Options& options,
...
-  TestRandomCaps(options, QueryType::CAP);
+  TestRandomCaps(options, index_type, query_type);
 }
 
-TEST(S2RegionTermIndexer, IndexRegionsQueryRegionsOptimizeSpace) {
-  S2RegionTermIndexer::Options options;
-  options.set_optimize_for_space(true);        // Optimize for space.
+TEST_P(S2RegionTermIndexerTest, UseLeafCells) {
   options.set_min_level(4);
-  options.set_max_level(S2CellId::kMaxLevel);  // Use leaf cells.
```
jmr, line 149: If you don't need to change something, don't change it.

src/s2/s2region_term_indexer_test.cc
```diff
@@ -44,10 +45,13 @@ S2_DEFINE_int32(iters, 400, "number of iterations for testing");
 
 namespace {
 
-enum QueryType { POINT, CAP };
+enum DataType {
+  POINT = 0,
```
jmr, line 49: Leave it like before. Explicit = 0 not required.

jmr (changes requested):

src/s2/s2region_term_indexer.cc
MBkkt: > Every allocation cost at least 10ns (on practice more), so if we make 1 instead of N it decrease overall time at least on (N-1)*10ns

It's significant for me not because it's big but because it costs nothing to me, I just should use other API
Even small improvements still good, if don't make code significantly more difficult

src/s2/s2region_term_indexer_test.cc
MBkkt: @jmr Do you ok with current tests?

src/s2/s2region_term_indexer.cc
```diff
@@ -178,14 +199,20 @@ vector<string> S2RegionTermIndexer::GetIndexTermsForCanonicalCovering(
...
-    if (level == true_max_level || !options_.optimize_for_space()) {
-      // Add an ancestor term for this cell at the constrained level.
-      terms.push_back(GetTerm(TermType::ANCESTOR, id.parent(level), prefix));
+    const bool is_max_level_cell = level == true_max_level;
+    // Add a term for this cell, max_level cell ANCESTOR is optimization
+    terms->push_back(GetTerm(is_max_level_cell ? TermType::ANCESTOR
+                                               : TermType::COVERING,
+                             id, prefix));
+
+    // If query only contains points, there are no need other terms.
```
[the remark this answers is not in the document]
MBkkt: Same as for index_only optimization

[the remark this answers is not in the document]
MBkkt: Do you want I remove them both, or only my comment?

src/s2/s2region_term_indexer_test.cc
MBkkt: >Use something more descriptive than "2" to differentiate it.

What? No good name was in old code

src/s2/s2region_term_indexer_test.cc
MBkkt: Ok

src/s2/s2region_term_indexer_test.cc
MBkkt: It's in test name now! It wasn't so I don't think it's needed anymore

src/s2/s2region_term_indexer_test.cc
MBkkt: I can use UseFaceCellsLevelMode2?

src/s2/s2region_term_indexer_test.cc
MBkkt: Done

src/s2/s2region_term_indexer_test.cc
MBkkt: Add some more descriptive

src/s2/s2region_term_indexer.cc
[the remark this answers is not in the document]
MBkkt: Removed

src/s2/s2region_term_indexer.cc
MBkkt: Concrete for me for query it will be same.
I can only see the difference only for indexing, but I thought it's better to have a symmetrical API.

Maybe as a compromise I can keep the new indexing API but remove the new querying API (if you think that would be better)

MBkkt:
@jmr Any problem here?

src/s2/s2region_term_indexer.h
```diff
@@ -289,6 +303,21 @@ class S2RegionTermIndexer {
   std::vector<std::string> GetQueryTermsForCanonicalCovering(
       const S2CellUnion& covering, absl::string_view prefix);
 
+  // Same as above but allows to reuse same buffer for different points or use
```
smcallis, line 306: I think its fine to have these since the old ones call the new variants.  Since the terms vector isn't optional let's just make it a non-const reference though.

src/s2/s2region_term_indexer.cc
```diff
@@ -61,6 +61,12 @@
 //   never be any document regions larger than the query region.  This can
 //   significantly reduce the size of queries.
 //
+// + If the query will contain only points (rather than general regions), then
+//   we can skip all the ancestor terms mentioned above (except last cell see
+//   `GetIndexTerms(const S2Point& point...` for details) because there will
+//   never be any document regions larger than the index region.  This can
```
smcallis, line 67: I think saying "never be any document regions smaller than the query region" makes the comment have the same sense as 1. above.

src/s2/s2region_term_indexer.h
smcallis: Ah yes if the performance gain isn't worth it then we should probably just elide.

MBkkt: > Are they cleared? appended to?

appended to, I think function is pretty small to see it from code.

> The formatting is also inconsistent (const T& vs const T &) with a few lines below. A lot of time is spent dealing with this for epsilon difference.

Sorry you don't have clang-format in repo, so my ide reformat it randomly, I tried to make it like was in other function, maybe in some places I missed it

> This seems like a small inefficiency we should forget about 97% of the time. @MBkkt even said the few microseconds it saved were irrelevant to him.

> This seems like a small inefficiency we should forget about 97% of the time. @MBkkt even said the few microseconds it saved were irrelevant to him.

https://github.com/google/s2geometry/pull/303#discussion_r1069346176

I already wrote, now it's ok for me don't have such api for query, but for indexing, calling other api make my code faster (a little but measurable), so I want other api.



### Offtop:

Btw today we discussed how speedup geo terms search: now we need to make disjunction of all query terms (it's from s2 documentation and it works now for us).
But it's possible to make conjuntion of disjunction which is a lot of better for inverted index (don't need to read a lot of postings lists)

Like query covering:
```
cellid-0: min_level 
cellid-1: min_level + 1
cellid-2: min_level + 1
```
Like query terms:
```
cellid-0-term-min-level
cellid-1-term-min-level
cellid-1-term-min-level+1
// cellid-2-term-min-level, for an example same as cellid-1-term-min-level, so deduplicated
cellid-2-term-min-level+1
```
So we can make query like:
```
(cellid-0-term-min-level || cellid-1-term-min-level) && (cellid-1-term-min-level+1 || cellid-2-term-min-level+1)
```
instead of 
```
cellid-0-term-min-level || cellid-1-term-min-level || cellid-1-term-min-level+1 || cellid-2-term-min-level+1
```

It has few other tricks, like we should store in index for pair: term, document in posting list: is it last or not

But in general it's a lot of better.

Why do I write about this?

Because for use such api we need to know term level.
It can be computed from terms, but it's not really nice, because we already know they from covering.

So will be nice to have overload in future like:

std::vector<std::pair<int/*level*/,std::string/*term*/>>

src/s2/s2region_term_indexer.cc
MBkkt: you're right

src/s2/s2region_term_indexer.h
MBkkt: > I think its fine to have these since the old ones call the new variants. Since the terms vector isn't optional let's just make it a non-const reference though.

@smcallis 

reference looks better for me, but in all other places in s2 mutable reference arguments passed like pointer

src/s2/s2region_term_indexer.cc
jmr: > I already wrote, now it's ok for me don't have such api for query, but for indexing, calling other api make my code faster (a little but measurable), so I want other api.

Do you have an indexing benchmark, or are you referring to the microbenchmarks you wrote in #304? What are the results?

> Offtop:
> Btw today we discussed how speedup geo terms search: now we need to make disjunction of all query terms (it's from s2 documentation and it works now for us).

As you say, off-topic for this PR. Please open a discussion or an issue.

MBkkt: @jmr Today another day when I thought that will be good to have such API

But because now you don't have it I need to write code like this:
```cpp
    geoTerms = _indexer.GetIndexTerms(*_shape.region(), {});
    if (!_shape.contains(centroid)) {
      auto terms = _indexer.GetIndexTerms(centroid, {});
      geoTerms.insert(geoTerms.end(), std::make_move_iterator(terms.begin()),
                      std::make_move_iterator(terms.end()));
```

MBkkt: > Do you have an indexing benchmark, or are you referring to the microbenchmarks you wrote in https://github.com/google/s2geometry/pull/304? What are the results?

We already computed it save at least N*10 ns, where N number of inserts.
Why do you against this API so much?

PR #312, opened 2023-04-24

jmr:
* Use absl::flags and absl::log instead of gflags/glog
* Use ABSL_UNREACHABLE, requires https://github.com/abseil/abseil-cpp/commit/6a87605
* New edge/circle intersection ordering predicates
* New functions `GetUVCoordOfEdge` and `GetIJCoordOfEdge`
* S2PolylineAlignment: Use Norm instead of Norm2
* S2Polygon::DestructiveUnion: Use priority_queue internally
* Update CONTRIBUTING.md with more detailed instructions
* Replace `const char *` with `absl::string_view`
* Replace most `gtl::dense_hash_set` uses with `absl::flat_hash_set`
* Replace most `std::set` and `std::map` with `absl::flat_hash_set`
  and `flat_hash_map`


jmr:
@MBkkt

src/s2/s2cell_id_test.cc
```diff
@@ -367,7 +368,7 @@ TEST(S2CellId, LegacyCoder) {
 
 TEST(S2CellId, LegacyCoderTokenInvalid) {
   // Token too long
-  absl::string_view invalid_token = "000000000000000404\0";
+  string_view invalid_token = "000000000000000404\0";
```
MBkkt, line 371: Unnecessary \0 at the end?

src/s2/s2edge_crossings_test.cc
```diff
@@ -383,12 +384,14 @@ class GetIntersectionStats {
...
     for (int i = 0; i < kNumMethods; ++i) {
       if (tally_[i] == 0) continue;
-      printf("%10s %9d %5.1f%% %9d %5.1f%%  %5.1f%%\n",
-             S2::internal::GetIntersectionMethodName(
-                 static_cast<IntersectionMethod>(i)),
-             tally_[i], 100.0 * tally_[i] / total,
-             totals[i], 100.0 * totals[i] / total,
-             100.0 * tally_[i] / totals[i]);
+      // Use `StrFormat` to handle `string_view`.
+      puts(absl::StrFormat("%10s %9d %5.1f%% %9d %5.1f%%  %5.1f%%",
```
MBkkt, line 388: Why puts? Why not `absl::PrintF()`

src/s2/s2polygon.cc
```diff
@@ -1282,41 +1291,48 @@ unique_ptr<S2Polygon> S2Polygon::DestructiveApproxUnion(
...
+  // Repeatedly union the two smallest polygons and add the result to the
+  // queue until we have a single polygon to return.  We use the index
+  // into `polygons` as the second component of the pair instead of a
+  // pointer so the comparison is deterministic rather than depending
+  // on the polygon's address.  (Some clients depend on determinism.)
+  using Pair = pair<int, size_t>;  // (num vertices, `polygons` index)
+  using PriorityQueue = priority_queue<Pair, vector<Pair>, greater<Pair>>;
+  PriorityQueue queue;  // Least vertices is `top`.
+  for (size_t i = 0; i < polygons.size(); ++i) {
+    queue.emplace(polygons[i]->num_vertices(), i);
```
MBkkt, line 1309: Size of `polygons` doesn't change here, so you can use address instead of compute it in loop with index

src/s2/s2polygon.cc
```diff
@@ -1282,41 +1291,48 @@ unique_ptr<S2Polygon> S2Polygon::DestructiveApproxUnion(
...
+    // Replace `a` with the union, and clear `b` to reclaim memory.
+    polygons[a.second] = std::move(union_polygon);
+    polygons[b.second].reset();
+
     // We assume that the number of vertices in the union polygon is the
     // sum of the number of vertices in the original polygons, which is not
     // always true, but will almost always be a decent approximation, and
     // faster than recomputing.
+    queue.emplace(a.first + b.first, a.second);
   }
```
MBkkt, line 1333: I see it old code, but is num_vertices slow for polygon? I thought it's computed in ctor

src/s2/s2shapeutil_shape_edge_id.h
```diff
@@ -44,13 +43,14 @@ struct ShapeEdgeId {
   bool operator>(ShapeEdgeId other) const;
   bool operator<=(ShapeEdgeId other) const;
   bool operator>=(ShapeEdgeId other) const;
+
+  template <typename H>
+  friend H AbslHashValue(H h, ShapeEdgeId id) {
+    return H::combine(std::move(h), id.shape_id, id.edge_id);
```
MBkkt, line 49: 👍

CMakeLists.txt
```diff
@@ -52,26 +43,8 @@ add_feature_info(S2_USE_SYSTEM_INCLUDES S2_USE_SYSTEM_INCLUDES
...
-
-if (WITH_GFLAGS)
-    find_package(GFlags REQUIRED)
-    if (NOT ${GFLAGS_FOUND})
-        message(FATAL_ERROR "GFlags requested but not found")
-    endif()
-    add_definitions(-DS2_USE_GFLAGS)
-endif()
+# Don't output anything for LOG(INFO).
+add_definitions(-DABSL_MIN_LOG_LEVEL=1)
```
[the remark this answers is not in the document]
jmr: Is that going to be a problem for you?

The main reason there was `S2_USE_GFLAGS` was in case someone didn't have it installed.  Since it's all part of absl, it's less important now, but I can imagine people wanting to swap in another flags/logging package or save the size.

I think `-DABSL_MIN_LOG_LEVEL=99 -DDABSL_MAX_VLOG_VERBOSITY=0` will disable logging. I don't think there's currently an "absl flags are just constants" option.

src/s2/s2polygon.cc
jmr: That makes the resulting polygon depend on the addresses of the input polygons, which some clients complained about.

CMakeLists.txt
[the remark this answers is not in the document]
MBkkt: I don't know, just noticed it can be big behavior change

src/s2/s2edge_crossings_test.cc
jmr: Thanks. I didn't know it existed.  This will be fixed in a future update.

src/s2/s2polygon.cc
MBkkt: Hmm I probably should looked how union polygons work.
Ok

src/s2/s2polygon.cc
jmr: Good catch. It looks like `num_vertices` changed back in 2009, so this is out of date.

src/s2/s2polygon.cc
MBkkt: Hmm I understand, above exist comment, I missed it

MBkkt: But wait, addresses of polygon in vector is monotonically increasing.
So I think this comment was about elements itself (addresses of polygons, not their cell in vector)

jmr: I see. Yeah, the comment was about `S2Polygon *`. You want to use a `unique_ptr<S2Polygon> *`?  Yeah, that's probably better since it can't be confused with num_vertices.

src/s2/s2polygon.cc
jmr: This will be fixed separately.

src/s2/s2cell.h
```diff
@@ -45,6 +47,26 @@ class S2LatLngRect;
...
+  // The boundaries are defined in UV coordinates.  The orientation may be
+  // rotated relative to other face cells, but are consistent within a face
+  // (i.e. a cell's left edge is its left-ward neighbor's right edge).
+  //
+  enum Boundary {
+    // clang-format off
+    kBottomEdge = 0,
+    kRightEdge  = 1,
+    kTopEdge    = 2,
+    kLeftEdge   = 3
```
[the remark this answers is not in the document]
jmr: clang-format also doesn't like the aligned `=`s, but we'll do something here. Thanks.

src/s2/s2cell_id_test.cc
jmr: Thanks.  This is pre-existing and will be fixed separately.

src/s2/s2coords.h
```diff
@@ -113,6 +114,13 @@
 // dimension of the surface, not the space it is embedded in).
 namespace S2 {
 
+// The maximum absolute error in U/V coordinates when converting from XYZ.
+//
+// The XYZ -> UV conversion is a single division per coordinate, which is
+// promised to be at most 0.5*DBL_EPSILON absolute error for values with
+// magnitude less than two.
+constexpr double kMaxXYZtoUVError = 0.5 * DBL_EPSILON;
```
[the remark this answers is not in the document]
jmr: There's an implicit TODO for all these C++17 features when we bump the minimum version.

src/s2/s2polygon.cc
jmr: We'll probably do this in a future release.

src/s2/s2polyline_alignment_test.cc
```diff
@@ -458,6 +451,27 @@ TEST(S2PolylineAlignmentTest, ExactHeaderFileExample) {
   VerifyCost(*a, *b);
 }
 
+TEST(S2PolylineAlignmentTest, DifferentPathForDistanceVersusSquaredDistance) {
+  // Tests that we get the correct path in the case where we have polylines at
+  // right angles, that would get a different matching of points for distance
+  // cost versus squared distance cost. If we had used squared distance for the
+  // cost the path would be {{0, 0}, {1, 0}, {2, 0}, {3, 1}, {3, 2}}; See
+  // https://screenshot.googleplex.com/7eeMjdSc5HeSeTD for the costs between the
```
rsned, line 459: Should probably add a check for internal URL's for screenshot.googleplex.com

src/s2/s2predicates.cc
```diff
@@ -789,9 +789,256 @@ int CompareEdgePairDistance(const S2Point& a0, const S2Point& a1,
...
+                                     ToLD(m), ToLD(n));
+    if (ans != 0) {
+      return ans;
+    }
+  }
+
+  return ExactIntersectionOrdering(  //
+      ToExact(a), ToExact(b), ToExact(c), ToExact(d), ToExact(m), ToExact(n));
+
+  return 0;
```
rsned, line 916: There two returns here (line 913 && 916)

smcallis: Shocked the linter doesn't catch this, I'll get it fixed for the next roll.  Glad to see you're still contributing btw =D

PR #313, opened 2023-04-24
Add ability to create Python wheels

selimnairb:
This pull request enables the creation of Python wheels via integration with CMake. This also changes the name of the Python library from `pywraps2` to the more ergonomic `s2geometry`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,9 +202,8 @@ python setup.py bdist_wheel
 
 The resulting wheel will be in the `dist` directory.
 
-> If OpenSSL is in a non-standard location (e.g., as installed by Homebrew on
-> macOS) make sure to set `OPENSSL_ROOT_DIR` when calling `setup.py`, for 
-> example: `OPENSSL_ROOT_DIR=/opt/homebrew/Cellar/openssl@3/3.1.0 python setup.py bdist_wheel`.
+> If OpenSSL is in a non-standard location make sure to set `OPENSSL_ROOT_DIR` 
+> when calling `setup.py`, see above for more information.
 
 ## Other S2 implementations
 
--- a/README.md
+++ b/README.md
@@ -122,6 +122,12 @@ Disable building of shared libraries with `-DBUILD_SHARED_LIBS=OFF`.
 
 Enable the python interface with `-DWITH_PYTHON=ON`.
 
+If OpenSSL is installed in a non-standard location set `OPENSSL_ROOT_DIR`
+before running configure, for example on macOS:
+```
+OPENSSL_ROOT_DIR=/opt/homebrew/Cellar/openssl@3/3.1.0 cmake -DCMAKE_PREFIX_PATH=/opt/homebrew -DCMAKE_CXX_STANDARD=17
+```
+
 ## Installing
 
 From `build` subdirectory:
@@ -179,6 +185,15 @@ Version 4.0 is required, but it should be easy to make it work 3.0 or probably
 even 2.0.
 
 Python 3 is required.
+
+### Creating wheels
+```
+python setup.py bdist_wheel
+```
+
+> If OpenSSL is in a non-standard location (e.g., as installed by Homebrew on
+> macOS) make sure to set `OPENSSL_ROOT_DIR` when calling `setup.py`, for 
+> example: `OPENSSL_ROOT_DIR=/opt/homebrew/Cellar/openssl@3/3.1.0 python setup.py bdist_wheel`.
 
 ## Other S2 implementations
 
--- a/README.md
+++ b/README.md
@@ -187,9 +187,20 @@ even 2.0.
 Python 3 is required.
 
 ### Creating wheels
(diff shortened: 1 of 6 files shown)


setup.cfg
```diff
@@ -0,0 +1,19 @@
+[metadata]
+name = s2geometry
+version = 0.10.0
+description = Python packaging of s2geometry
+author = Brian Miles
+author_email = selimnairb@gmail.com
+license= MIT
```
jmr, line 7: license is Apache-2.0, but it looks like using `classifiers` is recommended:

setup.cfg
```diff
@@ -0,0 +1,19 @@
...
+project_urls =
+    Source  = https://github.com/selimnairb/s2geometry
```
jmr, line 9: https://github.com/google/s2geometry

setup.py
```diff
@@ -0,0 +1,37 @@
...
+from pathlib import Path
+
+import cmake_build_extension
+import setuptools
+
+# This example is compliant with PEP517 and PEP518. It uses the setup.cfg file to store
+# most of the package metadata. However, build extensions are not supported and must be
+# configured in the setup.py.
+setuptools.setup(
+    # The resulting "mymath" archive contains two packages: mymath_swig and mymath_pybind.
```
jmr, line 11: delete comments that are copied from the template

src/python/CMakeLists.txt
```diff
@@ -1,21 +1,42 @@
+# Set SWIG policies
+cmake_policy(SET CMP0078 NEW)
```
jmr, line 2: what are these?

README.md
```diff
@@ -180,6 +186,26 @@ even 2.0.
...
+```
+
+Then build the wheel:
+```
+python setup.py bdist_wheel
+```
+
+The resulting wheel will be in the `dist` directory.
+
+> If OpenSSL is in a non-standard location (e.g., as installed by Homebrew on
```
jmr, line 205: Maybe a "see above" instead of repeating this.

jmr (changes requested):
Thanks. This is requested a lot and I never had time to look into it.

Note that we plan to delete the swig wrappers soon and use pybind11 or pyclif with a more Pythonic API.

setup.cfg
```diff
@@ -0,0 +1,22 @@
+[metadata]
+name = s2geometry
+version = 0.10.0
```
jmr, line 3: What's the allowed format of version? Can this be 0.11.dev or prerelease or something?

0.10 has already been released, so it's probably not great if we end up with loads of variants built at different times claiming to be 0.10.0.

src/python/CMakeLists.txt
jmr: Something a bit more informative like

```
# Generate standard target names.
cmake_policy(...)
# Honor SWIG_MODULE_NAME via -module flag.
cmake_policy(...)
```

selimnairb: These are documented in the [UseSWIG CMake docs](https://cmake.org/cmake/help/latest/module/UseSWIG.html):

## CMP0078:

[swig_add_library](https://cmake.org/cmake/help/latest/module/UseSWIG.html#command:swig_add_library):
Changed in version 3.13: This command creates a target with the specified <name> when policy [CMP0078](https://cmake.org/cmake/help/latest/policy/CMP0078.html#policy:CMP0078) is set to NEW. Otherwise, the legacy behavior will choose a different target name and store it in the SWIG_MODULE_<name>_REAL_NAME variable.

## CMP0086:

[Properties on Source Files](https://cmake.org/cmake/help/latest/module/UseSWIG.html#properties-on-source-files):

SWIG_MODULE_NAME
Specify the actual import name of the module in the target language. This is required if it cannot be scanned automatically from source or different from the module file basename. For example:

set_property(SOURCE mymod.i PROPERTY SWIG_MODULE_NAME mymod_realname)
Changed in version 3.14: If policy [CMP0086](https://cmake.org/cmake/help/latest/policy/CMP0086.html#policy:CMP0086) is set to NEW, -module <module_name> is passed to SWIG compiler.


I can't get the build to work without setting CMP0078 and CMP0086 to new. Since you are planning to migrate away from SWIG, I would recommend leaving this as is for now as they will be removed as part of that migration.

src/python/CMakeLists.txt
```diff
@@ -1,21 +1,42 @@
...
+cmake_policy(SET CMP0086 NEW)
+
+# Handle where to install the resulting Python package
+if(CALL_FROM_SETUP_PY)
```
jmr, line 6: `if (CALL_FROM_SETUP_PY)` for consistency

selimnairb:
@jmr Let me know if there's anything more to be done on this PR and if there is anyway that I can help with the migration from SWIG. Thanks!

jmr:
Thanks! I'll let you know when we have something in the way of a SWIG replacement.

PR #333, opened 2023-11-21
Add a CMake option BUILD_TESTS to decide whether to build S2 unit tests

acelyc111:
In the use case of using s2geometry as a thirdparty library of a project, it's not needed to build S2 unit tests, but it's possible to use test utilities in src/s2/s2testing.
In this case, `s2testing` has to be exported, but unit tests are not necessary to build.
This patch adds an option `BUILD_TESTS` to decide whether to build S2 unit tests.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,7 +32,7 @@ add_feature_info(SHARED_LIBS BUILD_SHARED_LIBS
                  "builds shared libraries instead of static.")
 
 option(BUILD_EXAMPLES "Build s2 documentation examples." ON)
-option(BUILD_TESTS "Build s2 unittests." OFF)
+option(BUILD_TESTS "Build s2 unittests." ON)
 
 option(WITH_PYTHON "Add python interface" OFF)
 add_feature_info(PYTHON WITH_PYTHON "provides python interface to S2")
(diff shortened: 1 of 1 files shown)


smcallis:
Is the intention to not build the unit tests themselves or just not build the google test/google mock subdirectories?  I think this would only do the latter.

acelyc111:
> Is the intention to not build the unit tests themselves or just not build the google test/google mock subdirectories? I think this would only do the latter.

@smcallis Both of them will not be built. As the code shows in:
https://github.com/google/s2geometry/pull/333/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR441, the unit tests will not be added.

CMakeLists.txt
```diff
@@ -432,7 +433,7 @@ install(TARGETS ${S2_TARGETS}
         LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}")
 
 message("GOOGLETEST_ROOT: ${GOOGLETEST_ROOT}")
-if (GOOGLETEST_ROOT)
+if (GOOGLETEST_ROOT AND BUILD_TESTS)
```
jmr, line 436: Depending on only `GOOGLETEST_ROOT` was a mistake.

How about something like
```
if (BUILD_TESTS)
  if (NOT GOOGLETEST_ROOT)
    message(FATAL_ERROR "BUILD_TESTS requires GOOGLETEST_ROOT")
  endif()
  message("GOOGLETEST_ROOT: ${GOOGLETEST_ROOT}")
  ...
```
?

acelyc111: Sure, done.

jmr:
Thanks!

PR #338, opened 2024-01-05
Fix several MSVC build issues

jherico:
Attempting to fix some of the issues preventing compilation on Windows platforms, in order to then fix the vcpkg version to also support windows, and finally to fix a downstream consumer of s2geometry so it can use vcpkg for the dependency.

### Changes

The `using Base::Base` technique to import the base class ctors doesn't appear to import the base class default ctor if it would otherwise be implicitly deleted (be the declaration of a non-default ctor in this case), at least on MSVC 2022.  See https://godbolt.org/z/EPnsMEq8v  Clang and GCC both seem to handle this OK.

`uint` isn't a primitive type or a guaranteed typedef in C++.  While Clang and GCC both appear to thread it as `unsigned int`, MSVC treats it as un unknown symbol.  Replacing it with `uint32_t` seems to do the trick.

### DIFF
--- a/src/s2/s2point.h
+++ b/src/s2/s2point.h
@@ -43,6 +43,25 @@ class S2Point : public Vector3_d {
   using Base = Vector3_d;
   using Base::Base;
 
+  // Due to an ambiguity in original C++11 specificiation, it was unclear
+  // whether imported base class default constructors should be considered
+  // when deciding to delete the default constructor of a class.  GCC and
+  // Clang both accept the base class default ctor, while MSVC 2017 and
+  // later do not.
+  // The ambiguity was resolved in
+  // https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/p0136r1.html
+  // and accepted as part of the C++17 approval process, with the intent of
+  // being retroactively applied to C++11.
+  // However, while MSVC accepted the change for C++17 and forward, it did
+  // not implement the change for prior versions. See their conformance page:
+  // https://learn.microsoft.com/en-us/cpp/overview/visual-cpp-language-conformance?view=msvc-170
+  // (search for P0136R1).
+  // The explicit declaration here of a default ctor is a workaround until
+  // this codebase is targeted at C++17 and above, at which point MSVC, GCC
+  // and Clang will all have identical behavior and won't require this
+  // explicit declaration of a default ctor.
+  S2Point() = default;
+
   // When S2Point was defined as a Vector3_d we could mix and match the two
   // names.  With inheritance upcasting to a Vector3_d is easy, but we need to
   // explicitly allow the other direction, even though there's no data to
--- a/src/s2/s1angle.cc
+++ b/src/s2/s1angle.cc
@@ -50,7 +50,7 @@ std::ostream& operator<<(std::ostream& os, S1Angle a) {
   char buffer[13];
   int sz = snprintf(buffer, sizeof(buffer), "%.7f", degrees);
   // Fix sign/unsign comparison for client that use `-Wextra` (e.g. Chrome).
-  if (sz >= 0 && static_cast<uint>(sz) < sizeof(buffer)) {
+  if (sz >= 0 && static_cast<uint32_t>(sz) < sizeof(buffer)) {
     return os << buffer;
   } else {
     return os << degrees;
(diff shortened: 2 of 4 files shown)


src/s2/s2point.h
```diff
@@ -43,6 +43,9 @@ class S2Point : public Vector3_d {
   using Base = Vector3_d;
   using Base::Base;
 
+  // This is required for building on MSVC
+  S2Point() = default;
```
jmr, line 47: Is this an MSVC bug?  I don't suppose they have a public bug tracker. Is there any link that could be added here?

src/s2/s2builder.h
```diff
@@ -1303,7 +1303,7 @@ inline void S2Builder::AddPoint(const S2Point& v) {
 }
 
 inline int S2Builder::num_input_edges() const {
-  return input_edges_.size();
+  return static_cast<int>(input_edges_.size());
```
jmr, line 1306: Is this to fix an error or warning?

jherico: that may have been a warning.  I can revert it.

src/s2/s2point.h
jherico: According to the spec, to the best of my ability to parse it, the `using Base::Base` declaration doesn't technically create new contractors in the derived class, but just makes the base class constructors available via overloading. So it's sort of left to the interpretation whether those imported "not quite constructors" count for the purposes of the rule that deletes the default ctor when a non-default ctor is defined.  GCC and Clang interpret it one way and MSVC interprets it another, and so far I can't find anything in the spec that's definitive one way or another.  

Personally I lean towards the MSVC interpretation, because if you add a data member to derived with a default value, and look at the GCC [code generated](https://godbolt.org/z/Gn7arTGsG) without an explicitly declared default ctor, it still includes code to initialize that derived member, even though the name of the ctor is `Derived::Base`.  Further, in my original example on compiler explorer, if you switch to the GCC output, the presence or absence of a `Derived() = default` declaration in the derived class produces the exact same output in terms of assembly.

It seems to me that the primary motivation in the `using Base::Base` declaration here (and in general) seems to be in avoiding having to re-declare a bunch of complex ctor signatures in a base class, not to circumvent the rules handling whether the default ctor is deleted.

So, the spec _may_ be ambiguous, and the presence of  `S2Point() = default;` _should_ be neutral on GCC and Clang and is required on MSVC.

src/s2/s2builder.h
jherico: verified it was only a warning.

src/s2/s2point.h
jherico: OK, so I asked on [Reddit](https://www.reddit.com/r/cpp_questions/comments/18zju2t/interaction_between_imported_base_class/kgi6crt/) and got a response that basically said this was ambiguous in the spec originally, but that the clarification was folded into the C++17 process and in theory should have been retroactively applied to C++11.

It appear that the MSVC compiler _does_ accept the code as written if you tell it the standard is C++17, but not at the current codebase level of C++14.

There are [guidelines](https://learn.microsoft.com/en-us/cpp/overview/how-to-report-a-problem-with-the-visual-cpp-toolset?view=msvc-170) for submitting a bug report for the MSVC toolset, but I'm already pretty far down the rabbit hole, since s2geometry is just a dependency of another project that I want to use as a dependency in my own project.

In order to get s2geometry working on MSVC without waiting on the turnaround on reporting the bug and waiting for a fix there are a couple options...

* Accept this change as is, since it should be innocuous with regard to GCC and Clang
* Wrap the change in a `#if __cplusplus >= 201703L || (defined(_MSC_VER) && _MSC_VER >= 1900)` macro

Additionally I can update the comment linking to either the reddit post or the linked paper.

jherico: Actually, looking at [this page](https://learn.microsoft.com/en-us/cpp/overview/visual-cpp-language-conformance?view=msvc-170) I suspect that reporting this as a bug isn't likely to get any traction, since the mention of complying with the reworded constructor inheritance paper is explicitly tagged as being something that has to be enabled via the `/std:C++17` argument.

src/s2/s2builder.h
jmr: Let's just keep this to fixing breakages.

src/s2/s2point.h
jmr: Can you put a summary of this in a comment and mention that it will no longer be required with `/std:C++17`?

src/s2/s2builder.h
jherico: I already reverted this in a prior commit, should have specified in my comment.

src/s2/s2point.h
jherico: commented.

jmr:
Thanks!

PR #339, opened 2024-01-10
Add modern cmake config targets to s2geometry

jherico:
The current vcpkg [version](https://github.com/microsoft/vcpkg/tree/master/ports/s2geometry) of s2geometry has to patch the [CMakeLists.txt](https://github.com/microsoft/vcpkg/blob/master/ports/s2geometry/CMakeLists.txt.patch) and generate a [Config.cmake.in](https://github.com/microsoft/vcpkg/blob/master/ports/s2geometry/Config.cmake.in.patch) file in order to create a modern CMake target file that can then be found by other packages via `find_package`.

Because this is a vcpkg addition, they mandate that this be placed in the `unofficial` namespace, in order to prevent any potential conflict with an eventual upstream config that might be introduced.  This PR introduces the upstream config so that it can be used by anyone who builds the package and so that the vcpkg version no longer requires any patching at all.

The changes here are essentially duplicating what the vcpkg patch does, except that they drop the `unofficial` prefix to everything, since the config would now be coming from the primary source.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -434,7 +434,10 @@ install(TARGETS ${S2_TARGETS}
         ARCHIVE DESTINATION "${CMAKE_INSTALL_LIBDIR}"
         LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}")
 
+# create an export type "s2Targets" detailing the targetable artifacts created by this project
 install(TARGETS s2 EXPORT s2Targets)
+# install the export targets as a CMake config file in the share/s2 folder so that they can referenced 
+# by downstream projects as `s2::s2` after a successful `find_package` call
 install(EXPORT s2Targets
         NAMESPACE s2
         FILE s2Targets.cmake
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -434,10 +434,12 @@ install(TARGETS ${S2_TARGETS}
         ARCHIVE DESTINATION "${CMAKE_INSTALL_LIBDIR}"
         LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}")
 
-# create an export type "s2Targets" detailing the targetable artifacts created by this project
+# Create an export type "s2Targets" detailing the targetable artifacts created
+# by this project.
 install(TARGETS s2 EXPORT s2Targets)
-# install the export targets as a CMake config file in the share/s2 folder so that they can referenced 
-# by downstream projects as `s2::s2` after a successful `find_package` call
+# Install the export targets as a CMake config file in the share/s2 folder so
+# that they can referenced  by downstream projects as `s2::s2` after a
+# successful `find_package` call.
 install(EXPORT s2Targets
         NAMESPACE s2
         FILE s2Targets.cmake
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -246,7 +246,10 @@ endif()
 if (S2_USE_SYSTEM_INCLUDES)
     target_include_directories(s2 SYSTEM PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/src)
 else ()
-    target_include_directories(s2 PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/src)
+    target_include_directories(s2 PUBLIC
+      $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src>
+      $<INSTALL_INTERFACE:include>
+    )
 endif ()
 
 # Add version information to the target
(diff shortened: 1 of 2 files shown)


jmr:
I will have to do some reading about modern CMake. I just based this on an example about ten years ago.

jmr (commented):
Does this change anything for people using it without vcpkg?

jherico:
> Does this change anything for people using it without vcpkg?

If you're asking if it will break existing downstream consumers, then the answer is no.  Downstream projects, regardless of how they're building, and whether or not they use the `S2_USE_SYSTEM_INCLUDES` option, should be unaffected by this change.

Strictly speaking it does change the build artifacts generated and installed (regardless of whether you're building from vcpkg or not), because it will create a few files in `<INSTALL_PREFIX>/share/s2`, in particular `s2Config.cmake` which contains the information required for downstream CMake builds to be to use s2 via [`find_package`](https://cmake.org/cmake/help/latest/command/find_package.html) in _Config_ mode as demonstrated here...

```CMake
find_package(s2 CONFIG REQUIRED)
target_link_library(my_project PUBLIC s2::s2)
```

This is the standard mechanism for packaging modern libraries built with CMake for other projects that are also using CMake.  Additionally the exported target `s2::s2` isn't simply a variable but a top-level CMake target type that contains properties, so the `target_link_library` doesn't just properly set the linker flags for the downstream project, but include directories and a variety of other possible [settings](https://cmake.org/cmake/help/latest/manual/cmake-properties.7.html#target-properties).  It also has the advantage of incorporating information about the distinct files required for the build type, so downstream consumers don't have to worry about matching release vs debug builds.

This makes lives easier for downstream consumers, with the caveat that package builders need to do a little more work to generate the config, but as you can see from this PR it's a _very_ small amount of work.

For further reading I'd suggest [this](https://cmake.org/cmake/help/latest/guide/importing-exporting/index.html#exporting-targets-from-the-build-tree) section of the CMake online documents, or a recent edition of [this book](https://crascit.com/professional-cmake/).  

I do kind of wish CMake would offer a simpler mode for developers that provided a prescribed file layout that _can_ be customized, but that makes writing a CMake file for a project like this one more trivial, the sort of "convention over configuration" approach Maven managed to do for Java, and which ultimately has become the norm for more modern languages like Rust, Go, Python, etc...  maybe someday.

CMakeLists.txt
```diff
@@ -246,7 +246,10 @@ endif()
 if (S2_USE_SYSTEM_INCLUDES)
     target_include_directories(s2 SYSTEM PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/src)
```
jmr, line 247: Should this be the same as the other branch just with `SYSTEM`?

CMakeLists.txt
```diff
@@ -432,6 +435,13 @@ install(TARGETS ${S2_TARGETS}
         ARCHIVE DESTINATION "${CMAKE_INSTALL_LIBDIR}"
         LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}")
 
+install(TARGETS s2 EXPORT s2Targets)
+install(EXPORT s2Targets
+  NAMESPACE s2
+  FILE s2Targets.cmake
+  DESTINATION share/s2/
```
jmr: Why `share/`? That seems specific to ... something.

CMakeLists.txt
```diff
@@ -432,6 +435,13 @@ install(TARGETS ${S2_TARGETS}
         ARCHIVE DESTINATION "${CMAKE_INSTALL_LIBDIR}"
         LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}")
 
+install(TARGETS s2 EXPORT s2Targets)
```
jmr, line 437: Nit: A high-level comment about what these are doing would be helpful.

jmr (changes requested):

CMakeLists.txt
jherico: Do you mean should line 247 be changed to this:

```
    target_include_directories(s2 SYSTEM PUBLIC
      $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src>
      $<INSTALL_INTERFACE:include>
    )
```

Then I would suggest the answer is "maybe, but probably no".  The addition of `S2_USE_SYSTEM_INCLUDES` probably means that someone is embedding this project inside their own project, perhaps as a git submodule, and then incorporating it into their project as a whole, meaning their resulting project builds s2 as well as their project. However, because the s2 includes are inside their own source tree, CMake will by default treat them as non-system include folders and generate warnings both during the build of s2 AND during the build of whatever downstream project includes the s2 headers.  If the downstream project has a policy of allowing zero warnings then they'd need this option to suppress them.  

The upshot is that anyone who is using this codepath is unlikely to be using the install target at all and I don't want to try to fix something that isn't currently broken.

CMakeLists.txt
jherico, line 437:
```suggestion
# create an export type "s2Targets" detailing the targetable artifacts created by this project
install(TARGETS s2 EXPORT s2Targets)
# install the export targets as a CMake config file in the share/s2 folder so that they can referenced 
# by downstream projects as `s2::s2` after a successful `find_package` call
```

jmr:
Thanks for the explanations, and sorry for the delay!

PR #354, opened 2024-02-29
abseil-cpp supports vlog

pateldeev:
abseil-cpp supports vlog in latest release. https://github.com/abseil/abseil-cpp/commit/7b6c17e378224844d8663a410da2da5353b205b9

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -227,6 +227,7 @@ target_link_libraries(
     absl::strings
     absl::type_traits
     absl::utility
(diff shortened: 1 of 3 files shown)


jmr:
Could you update README.md to say that https://github.com/abseil/abseil-cpp/releases/tag/20240116.1 is required?

Do we also need to bump the version to 0.12.0 in CMakeLists.txt?

pateldeev:
Updated readme. Add check in cmakes list that vlog module exists.

Issue #53, opened 2018-11-26
add bazel support


jmr:
We will need to do something about what are currently external dependencies: openssl, ~~glog, gflags~~.

This might be useful:
https://github.com/openssl/openssl/issues/3840

prem-nuro:
Maybe a dumb question, but could we just exclusively use boringssl with Bazel?
https://github.com/google/boringssl/tree/master-with-bazel

Then all deps already have Bazel support.

wyfSunflower:
1. @ericveach Could you please give me a solid reason why did S2 not use **_regular hexahedron_** projection, but use _**cube**_?
![image](https://user-images.githubusercontent.com/3376723/50130232-91ce5500-02b7-11e9-9e40-b18359cc7cca.png)

2. What's the distortion ratio and calculation time of  **_regular hexahedron_** for convert cell ids to points or vice versa, compare to **_cube_**?

chanced:
It seems openssl can be built with bazel using `rule_foreign_cc` https://github.com/bazelbuild/rules_foreign_cc/issues/337

I'm not familiar with bazel or c++ but I cruised down that rabbit hole.

chanced:
@wyfSunflower You can't use hexagons like that. You end up needing pentagons. If that's what you are after, go with [h3](https://github.com/uber/h3).

chrismgrayftsinc:
For what it's worth, it appears that boringssl is supported already.  See https://github.com/google/s2geometry/blob/5b5eccd54a08ae03b4467e79ffbb076d0b5f221e/src/s2/util/math/exactfloat/exactfloat.cc#L67 which appears to be the only place that OpenSSL / boringssl is used.

chrismgrayftsinc:
I created https://github.com/chrismgrayftsinc/s2geometry/tree/bazel as a proof of concept.  It compiles, but I have no idea if it works.  If a maintainer is interested in a PR, let me know and I can clean it up.

chrismgrayftsinc:
@jmr I use boringssl here: https://github.com/chrismgrayftsinc/s2geometry/commit/df42fefaf6aafe2b66ee072f7529a11cd7ec0bb7#diff-75c3ab7c9e0e33383d592e445ca6474cf69fc07110a420f0069c7e2e2798691dR14

In my reading, that's the only place where OpenSSL is used, but I could be wrong.

chrismgrayftsinc:
Ping

jmr:
[...]

PR #361, opened 2024-04-28
A bazel build and regression harness for s2geometry

spendres:
See issue #53. 

This version relies on the bazel central repository for abseil-cpp, googletest, and boringssl and points to HEAD of s2geometry.


src/s2/base/logging.h
```diff
@@ -0,0 +1,61 @@
+// Copyright Google Inc. All Rights Reserved.
```
jmr, line 1: This file should be deleted.

src/BUILD.bazel
```diff
@@ -0,0 +1,1962 @@
+# @s2geometry/BUILD.bazel
+load("@rules_cc//cc:defs.bzl", "cc_library", "cc_test")
+package(default_visibility = ["//visibility:public"])
+
+cc_library(name = "s2-includes",includes = ["s2"],)
```
jmr, line 5: what's this for? fix the formatting.

src/.bazelversion
```diff
@@ -0,0 +1 @@
+7.1.1
```
jmr, line 1: This looks like the latest release. What do we actually need?

src/README.md
```diff
@@ -0,0 +1,42 @@
+
+# S2 Geometry Library
```
jmr, line 2: This seems like a shortened version of https://github.com/google/s2geometry/blob/master/README.md
better to combine with that

src/BUILD.bazel
```diff
@@ -0,0 +1,1962 @@
...
+
+cc_library(
+    name = "base",
```
jmr, line 8: Can you explain the organization?  Why is this library here instead of in `src/s2/base/BUILD.bazel`?

jmr (changes requested):

src/README.md
spendres: 7.0.0 for bzlmod

src/BUILD.bazel
spendres: agreed

spendres:
...logging.h was removed in most recent commit/push.

spendres:
I tested build times for a hierarchical library definition as compared to a single flat library definition for s2. The flat definition builds twice as fast (12 sec vs 25 sec) about 25% of the time – using a mac M1. This is counter to expectations, so I reverted to the flat definition.

spendres:
Thanks for the requested changes by mmorel-35 … made and pushed.

jmr:
Thank you!

PR #372, opened 2024-07-04
Fix big endian build

rschu1ze:
PR https://github.com/google/s2geometry/pull/367 forgot to adjust a few types which are under the `IS_BIG_ENDIAN` define.


jmr:
Thanks!

PR #377, opened 2024-08-15
Create and install a cmake config version file

benbovy:
This will help handling s2 versions in downstream cmake projects.

`COMPATIBILITY SameMinorVersion` seems safer than `SameMajorVersion` or `AnyNewerVersion` (cf. some changes like S2ShapeIndex::CellRelation => S2CellRelation from v0.10 to v0.11).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -642,4 +642,12 @@ configure_package_config_file(${CMAKE_CURRENT_SOURCE_DIR}/Config.cmake.in
   INSTALL_DESTINATION "share/s2/"
   NO_SET_AND_CHECK_MACRO
   NO_CHECK_REQUIRED_COMPONENTS_MACRO)
-install(FILES ${CMAKE_CURRENT_BINARY_DIR}/s2Config.cmake DESTINATION "share/s2/")
+
+write_basic_package_version_file(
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -642,4 +642,12 @@ configure_package_config_file(${CMAKE_CURRENT_SOURCE_DIR}/Config.cmake.in
   INSTALL_DESTINATION "share/s2/"
   NO_SET_AND_CHECK_MACRO
   NO_CHECK_REQUIRED_COMPONENTS_MACRO)
-install(FILES ${CMAKE_CURRENT_BINARY_DIR}/s2Config.cmake DESTINATION "share/s2/")
+
+write_basic_package_version_file(
+  "${CMAKE_CURRENT_BINARY_DIR}/s2ConfigVersion.cmake"
+  VERSION ${PROJECT_VERSION}
+  COMPATIBILITY SameMinorVersion)
```
jmr, line 649: We want `SameMinorVersion` while we're at 0.x but then `SameMajorVersion` when we go to 1.x.  Is it going to cause any problems to change later?

benbovy: I guess it won't?

Another option may be to set compatibility to `AnyNewerVersion` and give users the freedom and responsibility to deal with breaking changes.

PR #378, opened 2024-08-20
Fix typos

mwtoews:
Typos were quickly identified and corrected using [codespell](https://github.com/codespell-project/codespell).

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,7 +138,7 @@ Prefix it with `sudo` if needed:
 sudo make install
 ```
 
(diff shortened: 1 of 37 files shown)


jmr:
Thanks!

jmr:
When I run `codespell`, I get a lot of false positives.  Did you run interactively and reject a bunch of corrections?

mwtoews:
yes, always interactive as false positives are anticipated

PR #379, opened 2024-09-18
Check for WIN32 rather than MSVC when building

asinghvi17:
This constraint is more general and occurs on all Windows platforms, where `MSVC` may not.  

xref https://github.com/JuliaPackaging/Yggdrasil/pull/9448, where we encountered this issue - the MSVC condition was not satisfied, but the build is on windows.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -70,7 +70,7 @@ if (WITH_PYTHON)
     find_package(Python3 COMPONENTS Interpreter Development)
 endif()
 
-if (MSVC)
+if (WIN32)
     # Use unsigned characters
     add_definitions(-J)
     # Make sure cmath header defines things like M_PI


jmr:
@asinghvi17 I can't merge this until you sign the CLA.

asinghvi17:
Hey, I'd signed the CLA a while ago - did it go through?  If the email address on this commit is causing issues I can also force-push to change it.

jmr:
If the email addresses don't match, it won't work.

https://github.com/google/s2geometry/pull/379/checks?check_run_id=30313549872

jmr:
@podsvirov @asinghvi17 when is `WIN32` needed?

podsvirov:
On Windows value WIN32 defined for all compilers (cl, gcc, clang...), but value MSVC only for cl.

CMakeLists.txt
```diff
@@ -70,7 +70,7 @@ if (WITH_PYTHON)
     find_package(Python3 COMPONENTS Interpreter Development)
 endif()
 
-if (MSVC)
+if (WIN32)
     # Use unsigned characters
     add_definitions(-J)
```
jmr, line 75: These are compiler options, so isn't `MSVC` right?  Do other compilers support the MSVC options? or maybe this should be `MSVC or something`?

jmr:
@asinghvi17 What's this `BinaryBuilder` environment mentioned in https://github.com/JuliaPackaging/Yggdrasil/pull/9448#issuecomment-2358199940 ?

CMakeLists.txt
jmr: Can you clarify this?

jmr:
Reopen if you have answers.

PR #388, opened 2024-11-04
apply PEP621

zacharyburnett:
PEP621 (https://peps.python.org/pep-0621/) defines using the `pyproject.toml` file to store project metadata in the `[project]` table. This has several advantages over using `setup.cfg` or `setup.py` to store the project metadata, namely:
- consolidates metadata and configuration into one file; `pyproject.toml` is required by PEP517 (https://peps.python.org/pep-0517/) and many third party tools use it to store their configuration
- allows alternate build tools other than `setuptools` (such as `hatch` or `meson`) to read the same project metadata


pyproject.toml
```diff
@@ -1,3 +1,26 @@
+[project]
+name = "s2geometry"
+version = "0.11.0.dev1"
```
zacharyburnett, line 3: we don't need to hardcode the version here
```suggestion
```

pyproject.toml
```diff
@@ -1,3 +1,26 @@
...
+description = "Python packaging of s2geometry"
```
zacharyburnett, line 4: use description from github repo
```suggestion
description = "Computational geometry and spatial indexing on the sphere"
```

pyproject.toml
```diff
@@ -6,3 +29,17 @@ requires = [
     "cmake_build_extension",
 ]
 build-backend = "setuptools.build_meta"
+
+[tool.setuptools]
+zip-safe = false
+include-package-data = false
+python-requres = ">=3.7"
```
zacharyburnett, line 36: move to `project.requires-python`
```suggestion
```

pyproject.toml
```diff
@@ -1,3 +1,26 @@
+[project]
+name = "s2geometry"
+version = "0.11.0.dev1"
+description = "Python packaging of s2geometry"
+authors = [
+    { name = "Brian Miles", email = "selimnairb@gmail.com" },
+]
+classifiers = [
```
zacharyburnett, line 8:
```suggestion
requires-python = ">=3.7"
classifiers = [
```

pyproject.toml
```diff
@@ -6,3 +34,21 @@ requires = [
...
+[tool.setuptools.packages.find]
+where = [
+    "src",
+]
+namespaces = false
+
+[tool.setuptools.package-dir]
+"" = "src"
+
+# [tool.pytest.ini-options]
```
jmr, line 51: Don't include commented out code.

pyproject.toml
```diff
@@ -1,3 +1,31 @@
...
+    { name = "Brian Miles", email = "selimnairb@gmail.com" },
+]
+requires-python = ">=3.7"
+classifiers = [
+    "Programming Language :: Python :: 3",
+    "Operating System :: POSIX",
+    "License :: OSI Approved :: Apache Software License",
+]
+dynamic = [
+    "version",
```
jmr, line 14: How is this different from the explicit version from `setup.cfg`?

pyproject.toml
```diff
@@ -1,3 +1,31 @@
+[project]
+name = "s2geometry"
+description = "Computational geometry and spatial indexing on the sphere"
+authors = [
+    { name = "Brian Miles", email = "selimnairb@gmail.com" },
```
jmr, line 5: Is this supposed to be the author of the project or the author of the `pyproject.toml` file?

jmr (changes requested):
Could you give a bit more info than just "apply PEP 621"? I don't know anything about Python packaging, so a few sentences of explanation would be helpful.

https://google.github.io/eng-practices/review/developer/cl-descriptions.html#informative

jmr:
@selimnairb This is moving the `setup.cfg` you added in #313 to `pyproject.toml`. Do you want to take a look?

pyproject.toml
zacharyburnett: this uses `setuptools_scm` to derive the project version from the Git tag (or use a `git describe`-like string for the version in between tagged commits). Just one less place to update the version when making a release

pyproject.toml
zacharyburnett: this should be the project author (you can include multiple)

selimnairb: @jmr Agreed, it shouldn't be me.

selimnairb:
> @selimnairb This is moving the `setup.cfg` you added in #313 to `pyproject.toml`. Do you want to take a look?

@jmr Yes, I'll pull it down today and have a look.

pyproject.toml
```diff
@@ -1,3 +1,35 @@
+[project]
+name = "s2geometry"
+description = "Computational geometry and spatial indexing on the sphere"
+authors = [
+    { name = "Dan Larkin-York", email = "dan@arangodb.com" },
+    { name = "Google Inc."},
+    { name = "Koordinates Limited"},
+    { name = "Mike Playle", email = "mike@mythik.co.uk" },
+    { name = "Tiago Brito", email = "tiago.brito@90poe.io" },
+]
```
zacharyburnett, line 10: how's this?

selimnairb: @jmr Overall, this looks good to me. Thanks @zacharyburnett this is a nice update to more current practice. @jmr Does the Python wheel get built and tested in CI? If not, I would recommend adding this (though not part of this PR). If so, it would be nice to see whether that works with this new method (I would imagine the CI script needs to be updated).

zacharyburnett: I'd recommend using `cibuildwheel` for that but you're right, out of scope of this PR

pyproject.toml
```diff
@@ -6,3 +38,21 @@ requires = [
...
+namespaces = false
+
+[tool.setuptools.package-dir]
+"" = "src"
+
+# [tool.pytest.ini-options]
+# testpaths = [
+#     "src/python/",
+# ]
```
zacharyburnett, line 58:
```suggestion
```

pyproject.toml
```diff
@@ -1,3 +1,35 @@
+[project]
+name = "s2geometry"
+description = "Computational geometry and spatial indexing on the sphere"
+authors = [
+    { name = "Dan Larkin-York", email = "dan@arangodb.com" },
```
jmr, line 5: Could you either use the contents of `CONTRIBUTORS` here or leave it like in `setup.cfg` (in which case I'll fix it later)?

pyproject.toml
```diff
@@ -1,3 +1,35 @@
...
+    "Programming Language :: Python :: 3",
+    "Operating System :: POSIX",
+    "License :: OSI Approved :: Apache Software License",
+]
+dynamic = [
+    "version",
+]
+
+[project.license]
+file = "LICENSE"
```
jmr, line 22: There's no way to use an SPDX identifier here?

CONTRIBUTORS
```diff
@@ -30,3 +30,4 @@ Mike Playle <mike@mythik.co.uk>
 Phil Elson <pelson.pub@gmail.com>
 Robert Coup <robert.coup@koordinates.com>
 Tiago Brito <tiago.brito@90poe.io>
+Zachary Burnett <zachary.r.burnett@gmail.com>
```
jmr, line 33: It looks like we were doing it wrong before, but I think you should also be in the AUTHORS file.

pyproject.toml
jmr: There's currently no CI. :(

pyproject.toml
zacharyburnett: sure, it can be 
```suggestion
text = "Apache 2"
```
but then, why have a license file? I personally prefer reducing the amount of duplication in configuration but we can definitely change it to be the above if you prefer

CONTRIBUTORS
jmr: You're still not in the AUTHORS file, but I'll fix it later with the others.

jmr (approved):

jmr:
Thanks!

PR #390, opened 2024-11-07
automatically build library with `cmake` and test with `ctest`

zacharyburnett:
run `cmake` and `make install` in GitHub Actions to test the build

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,41 @@
+name: build
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+  release:
+    types:
+      - released
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
+  cancel-in-progress: true
+
+permissions: read-all
+
+jobs:
+  cmake:
+    strategy:
+      matrix:
+        runs-on:
+          - ubuntu-latest
+          # macos-13 is an intel runner, macos-14 is apple silicon
+          - macos-13
+          - macos-latest
+      fail-fast: false
+    runs-on: ${{ matrix.runs-on }}
+    timeout-minutes: 30
+    steps:
+      - uses: zacharyburnett/setup-abseil-cpp@de39f445295c887839e30c864ffbbb1c0231bc83 # 1.0.5
+        with:
+          cmake-build-args: "-DCMAKE_CXX_STANDARD=17 -DABSL_PROPAGATE_CXX_STD=ON -DABSL_ENABLE_INSTALL=ON -DBUILD_TESTING=off -DCMAKE_POSITION_INDEPENDENT_CODE=ON"
+          abseil-version: "20240722.0"
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - run: mkdir build
+      - run: cmake -DCMAKE_CXX_STANDARD=17 -DCMAKE_PREFIX_PATH=/usr/local/ -DBUILD_TESTS=OFF ..
+        working-directory: build/
+      - run: sudo cmake --build . --parallel --target=install
+        working-directory: build/
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,53 @@
+name: test
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
+  cancel-in-progress: true
+
+permissions: read-all
+
+env:
+  GOOGLETEST_VERSION: "1.15.2"
+
+jobs:
+  ctest:
+    strategy:
+      matrix:
+        runs-on:
+          - ubuntu-latest
+          # macos-13 is an intel runner, macos-14 is apple silicon
+          - macos-13
+          - macos-latest
+      fail-fast: false
+    runs-on: ${{ matrix.runs-on }}
+    timeout-minutes: 30
+    env:
+      CTEST_OUTPUT_ON_FAILURE: ON
+    steps:
+      - uses: zacharyburnett/setup-abseil-cpp@de39f445295c887839e30c864ffbbb1c0231bc83 # 1.0.5
+        with:
+          cmake-build-args: "-DCMAKE_CXX_STANDARD=17 -DABSL_PROPAGATE_CXX_STD=ON -DABSL_ENABLE_INSTALL=ON -DBUILD_TESTING=off -DCMAKE_POSITION_INDEPENDENT_CODE=ON"
+          abseil-version: "20240722.0"
+      - name: retrieve googletest v${{ env.GOOGLETEST_VERSION }}
+        run: |
+          wget https://github.com/google/googletest/releases/download/v${{ env.GOOGLETEST_VERSION }}/googletest-${{ env.GOOGLETEST_VERSION }}.tar.gz
+          tar -xzvf googletest-${{ env.GOOGLETEST_VERSION }}.tar.gz
+          echo GOOGLETEST_ROOT=${{ runner.temp }}/googletest-${{ env.GOOGLETEST_VERSION }} >> $GITHUB_ENV
+        working-directory: ${{ runner.temp }}
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - run: mkdir build
+      - run: cmake -DCMAKE_CXX_STANDARD=17 -DCMAKE_PREFIX_PATH=/usr/local/ -DGOOGLETEST_ROOT=${{ env.GOOGLETEST_ROOT }} -DGOOGLETEST_VERSION=${{ env.GOOGLETEST_VERSION }} ..
+        working-directory: build/
+      - run: cmake --build . ${{ runner.os == 'macOS' && '--parallel' || '' }}
+        # building tests with `--parallel` is disabled on Linux because GitHub seems to automatically cancel the job due to resource limits(?) 
+        working-directory: build/
+      - if: always()
+        run: cmake --build . --parallel --target=test
+        working-directory: build/


jmr:
Thanks for trying this. I authorized and triggered the workflows.  They all fail.

https://github.com/google/s2geometry/actions/runs/11724915532/job/32662569678?pr=390

```console
Getting action download info
Error: Unable to resolve action `actions/checkout@v5`, unable to find version `v5`
```

Doesn't this also need to do something to install the dependencies like abseil-cpp?

zacharyburnett:
> Thanks for trying this. I authorized and triggered the workflows. They all fail.
> 
> https://github.com/google/s2geometry/actions/runs/11724915532/job/32662569678?pr=390
> 
> ```
> Getting action download info
> Error: Unable to resolve action `actions/checkout@v5`, unable to find version `v5`
> ```

oof, I always forget it only has `v4` 😅 I'll just pin it to a specific release

> 
> Doesn't this also need to do something to install the dependencies like abseil-cpp?

oh yes, I forgot about abseil

jmr:
> oof, I always forget it only has `v4` 😅 I'll just pin it to a specific release

Can you use `v4` instead of `v5`? Or not specify a version? Why is a hash better?

I triggered it again for another batch of failures.

zacharyburnett:
> > oof, I always forget it only has `v4` 😅 I'll just pin it to a specific release
> 
> Can you use `v4` instead of `v5`? Or not specify a version? Why is a hash better?
> 
> I triggered it again for another batch of failures.

it's better to use a commit hash for security, because if someone takes control of the upstream action repo they could theoretically retag an existing tag to point to malicious code

It's not a pressing issue though, so we could just change it to `v4` here, you're right

zacharyburnett:
thank you for being so patient as I iterate on this workflow!

jmr:
I ran it again.

zacharyburnett:
I'm able to run actions on my fork: https://github.com/zacharyburnett/s2geometry/actions

I'll try to get this running and fixed

.github/workflows/build.yml
```diff
@@ -0,0 +1,48 @@
...
+          - macos-latest
+          - windows-latest
+    runs-on: ${{ matrix.runs-on }}
+    steps:
+      - uses: jwlawson/actions-setup-cmake@802fa1a2c4e212495c05bf94dba2704a92a472be # v2.0.2
+        with:
+          cmake-version: ${{ inputs.cmake-version }}
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          repository: abseil/abseil-cpp
```
jmr, line 30: For now, this LTS should be used:
https://github.com/abseil/abseil-cpp/releases/tag/20240116.2

This will change when I do the next big update.

.github/workflows/build.yml
```diff
@@ -0,0 +1,48 @@
...
+          path: ./_abseil-cpp/
```
jmr, line 31: Is there any reason to prefix all paths with "./"?

jmr:
```console
Install the project...
-- Install configuration: ""
CMake Error at absl/base/cmake_install.cmake:46 (file):
  file cannot create directory: /usr/local/lib/pkgconfig.  Maybe need
  administrative privileges.
Call Stack (most recent call first):
  absl/cmake_install.cmake:47 (include)
  cmake_install.cmake:47 (include)
```

You probably need  ` -DCMAKE_INSTALL_PREFIX=...` for `abseil-cpp`.

I use `-DCMAKE_PREFIX_PATH=/tmp/absl-install-17 -DCMAKE_INSTALL_PREFIX=/tmp/absl-install-17 -DABSL_ENABLE_INSTALL=ON -DBUILD_TESTING=off -DCMAKE_VERBOSE_MAKEFILE=on -DCMAKE_POSITION_INDEPENDENT_CODE=ON -DCMAKE_CXX_STANDARD=17`, but I'm not sure how many of these are really needed.

Then build `s2geometry` with `-DCMAKE_PREFIX_PATH=/tmp/absl-install-17`.

zacharyburnett:
> ```
> Install the project...
> -- Install configuration: ""
> CMake Error at absl/base/cmake_install.cmake:46 (file):
>   file cannot create directory: /usr/local/lib/pkgconfig.  Maybe need
>   administrative privileges.
> Call Stack (most recent call first):
>   absl/cmake_install.cmake:47 (include)
>   cmake_install.cmake:47 (include)
> ```
> 
> You probably need ` -DCMAKE_INSTALL_PREFIX=...` for `abseil-cpp`.
> 
> I use `-DCMAKE_PREFIX_PATH=/tmp/absl-install-17 -DCMAKE_INSTALL_PREFIX=/tmp/absl-install-17 -DABSL_ENABLE_INSTALL=ON -DBUILD_TESTING=off -DCMAKE_VERBOSE_MAKEFILE=on -DCMAKE_POSITION_INDEPENDENT_CODE=ON -DCMAKE_CXX_STANDARD=17`, but I'm not sure how many of these are really needed.
> 
> Then build `s2geometry` with `-DCMAKE_PREFIX_PATH=/tmp/absl-install-17`.

ah, thank you! I apologize, I'm not familiar with building C++ at all, so I appreciate your help

zacharyburnett:
from testing on the fork, the `cmake` job seems to succeed building now:
[image](https://github.com/zacharyburnett/s2geometry/actions/runs/11747249530/job/32728776911)

jmr:
`nproc` doesn't exist on macOS.  `sysctl -n hw.physicalcpu` or `logicalcpu` is the equivalent.

.github/workflows/build.yml
```diff
@@ -0,0 +1,64 @@
...
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - uses: actions/setup-python@0b93645e9fea7318ecaed2b359559ac225c90a2b # v5.3.0
+        with:
+          python-version: ${{ matrix.python-version }}
+      - run: pip install build
+      - run: python -m build
+      - uses: actions/upload-artifact@b4b15b8c7c6ac21ea08fcf65892d2ee8f75cf882 # v4.4.3
+        with:
+          name: s2geometry-py${{ matrix.python-version }}-${{ runner.os }}
+          path: ./dist/*.whl
```
zacharyburnett, line 64: add this after getting access to the PyPI project (https://pypi.org/project/s2geometry/) and adding this workflow `build.yml` as a Trusted Publisher in the PyPI project settings (https://docs.pypi.org/trusted-publishers/). 
```suggestion
          path: ./dist/*.whl
      - if: (github.event_name == 'release') && (github.event.action == 'released')
        uses: pypa/gh-action-pypi-publish@15c56dba361d8335944d31a2ecd17d700fc7bcbc # v1.12.2
```

zacharyburnett:
it appears, after adding a workflow to test the wheels, the Python tests pass on Linux, but on macOS they fail on every Python version except for 3.13:  
[image](https://github.com/zacharyburnett/s2geometry/actions/runs/11783535854)

figroc:
@jmr @zacharyburnett I'd definitely transfer the s2geometry project on pypi to you as your needs. My forked hasn't been maintained for years. All I needed was the ability to install from pypi.

zacharyburnett:
> @jmr @zacharyburnett I'd definitely transfer the s2geometry project on pypi to you as your needs. My forked hasn't been maintained for years. All I needed was the ability to install from pypi.

Thank you @figroc! I'm currently trying to figure out why the macOS wheel segfaults instantly, but the Linux wheels look like they're good to go!

jmr:
For the ubuntu ctest failures:

https://github.com/google/s2geometry/actions/runs/11785427420/job/32828227326?pr=390#step:7:119

googletest 1.10 is very old (2019).  It looks like 1.11 will fix this "maybe uninitialized" error.  I'm using 1.14.  I should update the README, but I'm not sure what to recommend/require.

A newer version might also fix the macOS problems.

.github/workflows/test.yml
```diff
@@ -0,0 +1,70 @@
...
+          echo GOOGLETEST_ROOT=${{ runner.temp }}/googletest-${{ env.GOOGLETEST_VERSION }} >> $GITHUB_ENV
+        working-directory: ${{ runner.temp }}
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - run: mkdir build
+      - run: cmake -DCMAKE_CXX_STANDARD=17 -DCMAKE_PREFIX_PATH=/usr/local/ -DGOOGLETEST_ROOT=${{ env.GOOGLETEST_ROOT }} -DGOOGLETEST_VERSION=${{ env.GOOGLETEST_VERSION }} ..
+        working-directory: build/
+      - run: cmake --build . --parallel
+        working-directory: build/
+      - if: always()
+        run: cmake --build . --parallel --target=test
```
jmr, line 41: Can you set `CTEST_OUTPUT_ON_FAILURE` or use `ctest --output-on-failure`? (I've never actually used ctest directly.)  It's helpful to see why tests fail.

https://cmake.org/cmake/help/latest/envvar/CTEST_OUTPUT_ON_FAILURE.html#envvar:CTEST_OUTPUT_ON_FAILURE

zacharyburnett: I added the env var to https://github.com/zacharyburnett/s2geometry/actions/runs/11841933989/job/32999371748, and it seems to be producing more output. Is that helpful?

zacharyburnett:
It appears the Linux `ctest` job was being terminated automatically for using too many resources; I removed `--parallel` from the build and it seems to succeed and pass all of the tests now

jmr:
> It appears the Linux `ctest` job was being terminated automatically for using too many resources; 

What was the error message? Do you have a link to the failure?

> I removed `--parallel` from the build and it seems to succeed and pass all of the tests now. 

Can you add a comment why `--parallel` has been disabled?  13 minutes is not ideal.

> `pytest` still fails on macOS with a segmentation fault, however

Maybe just remove these with a comment. I think we don't want to get in the habit of thinking failing checks are ok.

test/ctest (macos-latest) takes 140 minutes to fail.  That's way too slow.

I need to learn a bit about github actions and their security implications.  It will probably take me a week or two to be able to merge this.

.github/workflows/test.yml
jmr: Yes, thanks.

.gitignore
```diff
@@ -0,0 +1,3 @@
+dist/
```
jmr, line 1: Add a comment like
`# Python wheel directory`
(or whatever actually produces dist/).

setup.py
```diff
@@ -1,10 +1,13 @@
+import os
 import sys
 from pathlib import Path
 
 import cmake_build_extension
 import setuptools
 
 
+SETUPTOOLS_CMAKE_OPTIONS = os.environ.get("SETUPTOOLS_CMAKE_OPTIONS", "-DBUILD_TESTS=OFF").split()
```
jmr, line 9: What's using this?

zacharyburnett: I was using this to test whether inserting `GOOGLETEST_ROOT` into the cmake options would solve the segfault; I can remove it if desired

.github/workflows/build.yml
```diff
@@ -47,6 +47,28 @@ jobs:
           - "3.11"
           - "3.12"
           - "3.13"
+        exclude:
+          # importing the package from the wheel instantly segfaults on macOS Python <= 3.12
```
b4rdos, line 51: @zacharyburnett Have you found out why that's the case? I have been trying to build a wheel for macos arm64 for python 3.10. I can build the  wheel and install it and run it (`import s2geometry`) on any environment that shares the same base python. Eg venvs created using homebrew's 3.10.  But as soon as I switch to another installation (eg pyenv 3.10) I get segfaults. Something about KERN_INVALID_ADDRESS.

I used delocate-wheel to "fix" the wheel so it includes the necessary libraries as dylib. The resulting wheel segfaults even if the same base python was used. I noticed that Python.dylib is also linked and included.

b4rdos:
Is the idea for this to also publish to Pypi the built wheels? That would be awesome. I think that most recent one is 0.9.0

.github/workflows/build.yml
zacharyburnett: unfortunately, this is the limit of my knowledge of libraries and compilation 😓 

At the very least, we have working Python 3.13 wheels on macOS

zacharyburnett:
> Is the idea for this to also publish to Pypi the built wheels? That would be awesome. I think that most recent one is 0.9.0

Yes indeed, and to do so automatically on every GitHub release

.github/workflows/build.yml
b4rdos: Does this mean you are able to run a delocated wheel built with 3.13?

.gitignore
```diff
@@ -0,0 +1,3 @@
...
+*.egg-info
```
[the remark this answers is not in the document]
b4rdos: I think it's the `pip install .` step

.github/workflows/build.yml
```diff
@@ -0,0 +1,90 @@
...
+    steps:
+      - uses: zacharyburnett/setup-abseil-cpp@de39f445295c887839e30c864ffbbb1c0231bc83 # 1.0.5
+        with:
+          cmake-build-args: "-DCMAKE_CXX_STANDARD=17 -DABSL_PROPAGATE_CXX_STD=ON -DABSL_ENABLE_INSTALL=ON -DBUILD_TESTING=off -DCMAKE_POSITION_INDEPENDENT_CODE=ON"
+          abseil-version: "20240722.0"
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - uses: actions/setup-python@0b93645e9fea7318ecaed2b359559ac225c90a2b # v5.3.0
+        with:
+          python-version: ${{ matrix.python-version }}
+      - run: pip install build
```
b4rdos, line 83: Specifically for the macos wheel there will need to be a `delocate` step. What was the issue with cibuildwheel? Their step includes delocation. Also a test step that installs the wheel itself. Which is where I encountered issues.

b4rdos: This might be the case for other Linux based distros.

jmr: Why is `delocate` needed? If I understand correctly, it's to remove external shared libs. Which ones would it remove?

jmr:
So I should review all actions transitively depended on by this.  Is there an easy way to get a list?

I've seen people use [`manylinux`](https://github.com/pypa/manylinux) to build wheels. Would that be good to use?

.github/workflows/build.yml
b4rdos: It's not removing them, it's actually bundling them in the wheel and updates any references to point to the new location.

I have had issues sharing a wheel built on macos arm using a python from pyenv because there were harcoded paths with my username.

Searching for a solution pointed me to delocate.

zacharyburnett: as far as why I stopped using `pypa/cibuildwheel`, I found that it was difficult to include the installed `abseil-cpp` in `cibuildwheel`'s container, so I figured it would be easier to replicate the steps used in `cibuildwheel`, which includes `pypa/build`, instead: https://github.com/google/s2geometry/pull/390#issuecomment-2468585295

.github/workflows/build.yml
zacharyburnett: I didn't even have to `delocate` it, it just works 🤷

zacharyburnett:
> So I should review all actions transitively depended on by this. Is there an easy way to get a list?

Do you mean the dependencies of the actions used? As far as I know no, there's no easy way to get that list

> I've seen people use [`manylinux`](https://github.com/pypa/manylinux) to build wheels. Would that be good to use?

`cibuildwheel` uses a docker container to ensure `manylinux` compatibility; I'll look into that real quick

.github/workflows/build.yml
b4rdos: How did you test it works? Do you have a wheel you can share?

zacharyburnett: sure, here're the tests on `macos-13` (Intel) and `macos-latest` (ARM64): https://github.com/zacharyburnett/s2geometry/actions/runs/11935111543/job/33265728139

The tests fail with assert differences, but they actually run instead of segfaulting

Here're the built wheels: https://github.com/zacharyburnett/s2geometry/actions/runs/11935111540

b4rdos: The whole problem is being able to install the wheel on another machine with similar architecture. I have only been able to share a wheel I built by instructing them to only use python 3.10 and abseil from homebrew. (btw abseil from homebrew also works).

The above wheel will fail on my machine if try it on an environment built using pyenv's python.

zacharyburnett: Aha, you're right! In a new environment on a macOS Intel machine I get `Library not loaded`:
```
micromamba create -n s2py313 python=3.13 pytest -y
micromamba activate s2py313
pip install ~/Downloads/s2geometry-0.0.0-cp313-cp313-macosx_13_0_universal2.whl
pytest src/python/s2geometry_test.py
```
```
...
E     Reason: tried: '/Library/Frameworks/Python.framework/Versions/3.13/Python' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/Library/Frameworks/Python.framework/Versions/3.13/Python' (no such file), '/Library/Frameworks/Python.framework/Versions/3.13/Python' (no such file)
```

.github/workflows/build.yml
b4rdos: As an example, when I run `wheel unpack` on the built wheel and then `vim` the `_s2geometry.so` binary. Among the random symbols I see things like 
image

which are hardcoded paths that show specific dependencies on python and libssl install locations.

The result of `delocate-wheel` used by `cibuildwheel` is a wheel with `.dylibs` directory. The new `_s2geometry.so` has library references like 
image

.github/workflows/build.yml
zacharyburnett: oof, it gets a segfault when I `delocate-wheel` and then try running that on my machine, just as you said 😓

.github/workflows/build.yml
```diff
@@ -0,0 +1,91 @@
...
+        with:
+          cmake-build-args: "-DCMAKE_CXX_STANDARD=17 -DABSL_PROPAGATE_CXX_STD=ON -DABSL_ENABLE_INSTALL=ON -DBUILD_TESTING=off -DCMAKE_POSITION_INDEPENDENT_CODE=ON"
+          abseil-version: "20240722.0"
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - uses: actions/setup-python@0b93645e9fea7318ecaed2b359559ac225c90a2b # v5.3.0
+        with:
+          python-version: ${{ matrix.python-version }}
+      - run: pip install build delocate
+      - run: python -m build -w
+      - run: delocate-wheel ./dist/*.whl
```
b4rdos, line 85: FYI `delocate` is specific for macos. There is `auditwheel` for linux based.

A quick search pointed to me [repairwheel](https://github.com/jvolkman/repairwheel) as a wrapper for both.

.github/workflows/build.yml
jmr: I'm surprised there would be hardcoded absolute paths in the output, but ok.

zacharyburnett: It might be worth revisiting using the `cibuildwheel` action and attempting to insert `abseil-cpp` into the containers it spins up; as far as I can tell, `cibuildwheel` uses the `manylinux` container and also does macOS emulation but using `libc++` instead of `libSystem`

setup.py
jmr: Yeah, just revert this file unless you really need the changes.

.github/workflows/build.yml
```diff
@@ -0,0 +1,91 @@
...
+  pull_request:
+  push:
+    branches:
+      - master
+  release:
+    types:
+      - released
+  workflow_dispatch:
+
+jobs:
```
jmr, line 13: Prevent previous workflows from running.

```
concurrency:
  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
  cancel-in-progress: true
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,91 @@
+name: build
```
jmr, line 1: permissions: read-all

zacharyburnett:
so I've been QUITE dumb :sweat_smile: 

All this time I've been trying to recreate `cibuildwheel`'s functionality because I didn't know how to include the built version of `abseil-cpp` in the container, BUT I neglected to realize that I could have just been installing it with `yum` and `brew` (see https://github.com/google/s2geometry/pull/390#discussion_r1866392968). I'm testing with `cibuilwheel` now, with a `before-all` command to install the requisite libraries in the config

zacharyburnett:
taking this back into draft until I can figure out the magic config for `cibuildwheel` that'll work

jmr:
Ok, thanks for working so hard on this.

jmr:
I think getting some automated test running would still be a great improvement.  Could you just take out/disable the failing macOS parts?  That could be a separate PR when it's ready.

.github/workflows/build.yml
```diff
@@ -0,0 +1,41 @@
...
+        runs-on:
+          - ubuntu-latest
+          # macos-13 is an intel runner, macos-14 is apple silicon
+          - macos-13
+          - macos-latest
+      fail-fast: false
+    runs-on: ${{ matrix.runs-on }}
+    timeout-minutes: 30
+    steps:
+      - uses: zacharyburnett/setup-abseil-cpp@de39f445295c887839e30c864ffbbb1c0231bc83 # 1.0.5
```
jmr, line 32: Just wondering, what's the difference between `setup-abseil-cpp` and using `vcpkg` or `homebrew` to install `abseil-cpp`?  I see both `vcpkg` and `homebrew` are installed on the [runner images](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md).

jmr:
Thanks!

jmr:
macos-latest times out: https://github.com/google/s2geometry/actions/runs/13091503445/job/36528655440

This should not take 30 minutes, the other runners are much faster.

.github/workflows/build.yml
zacharyburnett: This was a while ago, but I think I remember that the `homebrew` recipe was not configured such that it could link properly, which is why I resorted to building it manually; I did not try `vcpkg` though

jmr:
This doesn't use `ctest` and I got some errors about not finding the binary when I tried to use it.  Do you know how to use ctest?

Also, is there any point in having separate build and test workflows?  We'll need to build everything for the tests anyway.

Issue #402, opened 2024-12-04
googletest version mismatch warning with bazel build


jmr:
```
% cd src
% bazel build //:all
WARNING: For repository 'googletest', the root module requires module version googletest@1.14.0, but got googletest@1.15.2 in the resolved dependency graph.
```

jmr:
@spendres

spendres:
The project builds and passes regression tests while using googletest@1.15.2 without out any source changes.

I will create a PR to update BUILD.bazel to use googletest@1.15.2 (latest).

PR #404, opened 2024-12-05
#402 update googletest for bazel build

spendres:
Updated bazel build to googletest@1.15.2.


jmr:
So if I see a warning like that, I should just update the version (and test that it still works)?

Issue #398, opened 2024-12-02
Unable to build wheel after #394 on macos arm64


b4rdos:
## Machine

```sh
$ uname -vm
Darwin Kernel Version 24.1.0: Thu Oct 10 21:02:45 PDT 2024; root:xnu-11215.41.3~2/RELEASE_ARM64_T8112 arm6
```

## Version

```sh
$ git rev-parse HEAD
778502a21eb9c6e813df72ce9518acfdfeb8e966
```

## Diff
```
$ git diff
diff --git a/setup.py b/setup.py
index 242882a..b352381 100644
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,8 @@ setuptools.setup(
                                         '-DCALL_FROM_SETUP_PY:BOOL=ON',
                                         '-DBUILD_SHARED_LIBS:BOOL=OFF',
                                         '-DCMAKE_POSITION_INDEPENDENT_CODE=ON',
-                                        '-DWITH_PYTHON=ON'
+                                        '-DWITH_PYTHON=ON',
+                                        '-DBUILD_TESTS=OFF',
                                     ]
         )
     ],
```

## Python
```sh
/opt/homebrew/bin/python3.13 -m venv brew313
source brew313/bin/activate
pip install cmake_build_extension wheel build -qq
python -m build .
```

## Error
```sh
ld: symbol(s) not found for architecture arm64
c++: error: linker command failed with exit code 1 (use -v to see invocation)
ninja: build stopped: subcommand failed.
```
---

I started going back commits using
```sh
git checkout HEAD~1
python -m build .
```

The version that produced a wheel successfully was #388. So something broke with #394 unless the approach is now different.

jmr:
@andyneff

andyneff:
I'm trying to look into this now. What is the output of `pip freeze` for your venv?

b4rdos:
> I'm trying to look into this now. What is the output of `pip freeze` for your venv?

```sh
❯ pip freeze
build==1.2.2.post1
cmake==3.31.1
cmake-build-extension==0.6.1
gitdb==4.0.11
GitPython==3.1.43
ninja==1.11.1.2
packaging==24.2
pyproject_hooks==1.2.0
setuptools==75.6.0
setuptools-scm==8.1.0
smmap==5.0.1
wheel==0.45.1
```

b4rdos:
Also, I have abseil installed via homebrew.
```sh
$ brew info abseil
[...]

PR #405, opened 2024-12-05
Suggested fix for linking

andyneff:
Better resolution to #398

Tested on Linux (in python and manylinux docker images), Windows 11 w/ VS 2022, and macOS

The [officially suggested fix](https://gitlab.kitware.com/cmake/cmake/-/issues/26503#note_1600940) works, and feels much better than my initial workaround.


jmr:
Thanks!

PR #406, opened 2025-01-26
build Python wheels with `cibuildwheel` and add to build workflow

### LINKED ISSUE #389: Use `cibuildwheel` for Python CI
I'd recommend using `cibuildwheel` for that but you're right, out of scope of this PR

_Originally posted by @zacharyburnett in https://github.com/google/s2geometry/pull/388#discussion_r1828232233_

zacharyburnett:
closes #389, ~~blocked by #390~~ [fixed]

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -95,6 +95,7 @@ jobs:
           //...
       working-directory: src
   python:
+    timeout-minutes: 30
     strategy:
       matrix:
         runs-on:
@@ -122,6 +123,7 @@ jobs:
   python-publish:
     if: (github.event_name == 'release') && (github.event.action == 'released')
     needs: [python]
+    timeout-minutes: 30
     runs-on: ubuntu-latest
     permissions:
       id-token: write
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -94,3 +94,51 @@ jobs:
           --test_verbose_timeout_warnings \
           //...
       working-directory: src
+  python:
+    strategy:
+      matrix:
+        runs-on:
+          - ubuntu-latest
+          # `large` is x86-64 and `xlarge` is AArch64.
+          - macos-latest-large
+          # macos-latest-xlarge may timeout.
+          # https://github.com/google/s2geometry/issues/409
+          - macos-latest-xlarge
+      fail-fast: false
+    runs-on: ${{ matrix.runs-on }}
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          fetch-tags: true
+          persist-credentials: false
+      - uses: pypa/cibuildwheel@8d2b08b68458a16aeb24b64e68a09ab1c8e82084 # v3.4.1
+        with:
+          package-dir: ./
+          output-dir: dist/
+      - uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # v7.0.0
+        with:
+          name: dist-${{ runner.os }}-${{ runner.arch }}
+          path: ./dist/
+  python-publish:
+    if: (github.event_name == 'release') && (github.event.action == 'released')
+    needs: [python]
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      attestations: write
+    # Requires environment protection rules in GitHub Settings:
+    # Settings > Environments > pypi > Add required reviewers
+    environment:
+      name: pypi
+      url: https://pypi.org/p/s2geometry
+    steps:
+      - uses: actions/download-artifact@37930b1c2abaa49bbe596cd826c3c89aef350131 # v7.0.0
+        with:
+          pattern: dist*
+          path: dist/
+          merge-multiple: true
+      - uses: actions/attest-build-provenance@96278af6caaf10aea03fd8d33a09a777ca52d62f # v3.2.0
+        with:
+          subject-path: "dist/*"
+      # To upload to PyPI without a token, add this workflow file as a Trusted Publisher in the project settings on the PyPI website
+      - uses: pypa/gh-action-pypi-publish@ed0c53931b1dc9bd32cbe73a98c7f6766f8a527e # v1.13.0
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,9 @@
 cmake_minimum_required(VERSION 3.18)
 
+# avoids error: aligned allocation function of type 'void *(std::size_t, std::align_val_t)' is only available on macOS 10.13 or newer
+# see https://github.com/google/oss-policies-info/blob/main/foundational-cxx-support-matrix.md#macos
+set(CMAKE_OSX_DEPLOYMENT_TARGET "10.13" CACHE STRING "Minimum macOS version")
+
 project(s2-geometry
         VERSION 0.12.0)
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -63,3 +63,7 @@ tag_regex = "^(?:[\\w-]+-)?(?P<version>[vV]?\\d+(?:\\.\\d+){0,2}[^\\+]*)(?:\\+.*
  
 [tool.cibuildwheel]
 test-command = "python -c 'import s2geometry'"
+
+[tool.cibuildwheel.macos]
+# skip wheel delocation (command can't find the binary)
+repair-wheel-command = ""
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,7 +32,6 @@ Source = "https://github.com/google/s2geometry"
 [dependency-groups]
 test = [
     "pytest",
-    "abi3audit"
 ]
 
 [build-system]
@@ -66,6 +65,7 @@ tag_regex = "^(?:[\\w-]+-)?(?P<version>[vV]?\\d+(?:\\.\\d+){0,2}[^\\+]*)(?:\\+.*
 # we only need to build with CPython 3.10 (because of the limited API / stable ABI)
 build = "cp310-*"
 test-groups = ["test"]
+test-requires = ["abi3audit"]
 test-command = "pytest {project}/src/python/s2geometry_test.py && abi3audit --strict --report {project}/dist/*.whl"
 
 [tool.cibuildwheel.macos]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,6 +32,7 @@ Source = "https://github.com/google/s2geometry"
 [dependency-groups]
 test = [
     "pytest",
+    "abi3audit"
 ]
 
 [build-system]
@@ -63,7 +64,7 @@ tag_regex = "^(?:[\\w-]+-)?(?P<version>[vV]?\\d+(?:\\.\\d+){0,2}[^\\+]*)(?:\\+.*
  
 [tool.cibuildwheel]
 test-groups = ["test"]
-test-command = "pytest {project}/src/python/s2geometry_test.py"
+test-command = "pytest {project}/src/python/s2geometry_test.py && abi3audit --strict --report {project}/dist/*.whl"
 
 [tool.cibuildwheel.macos]
 # skip wheel delocation (command can't find the binary)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -69,7 +69,6 @@ test-requires = ["abi3audit"]
 test-command = "pytest {project}/src/python/s2geometry_test.py && abi3audit --strict --report {project}/dist/*.whl"
 
 [tool.cibuildwheel.macos]
-# skip wheel delocation on macOS because of the following errors on both architectures:
-# delocate.libsana.DelocationError: Failed to find any binary with the required architecture: 'arm64'
-# delocate.libsana.DelocationError: Failed to find any binary with the required architecture: 'x86_64'
+# delocate is not needed because BUILD_SHARED_LIBS is OFF in setup.py,
+# so all dependencies are statically linked into the extension module.
 repair-wheel-command = ""
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -69,5 +69,7 @@ test-requires = ["abi3audit"]
 test-command = "pytest {project}/src/python/s2geometry_test.py && abi3audit --strict --report {project}/dist/*.whl"
 
 [tool.cibuildwheel.macos]
-# skip wheel delocation (command can't find the binary)
+# skip wheel delocation on macOS because of the following errors on both architectures:
+# delocate.libsana.DelocationError: Failed to find any binary with the required architecture: 'arm64'
+# delocate.libsana.DelocationError: Failed to find any binary with the required architecture: 'x86_64'
 repair-wheel-command = ""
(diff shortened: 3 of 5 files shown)


zacharyburnett:
The current blocking issue with building is that `repairwheel` can't find the C extension in the wheel:
```
Repairing wheel...

    + sh -c 'auditwheel repair -w /tmp/cibuildwheel/repaired_wheel /tmp/cibuildwheel/built_wheel/s2geometry-0.0.0-cp37-cp37m-linux_x86_64.whl'
INFO:auditwheel.main_repair:Repairing s2geometry-0.0.0-cp37-cp37m-linux_x86_64.whl
INFO:auditwheel.main_repair:This does not look like a platform wheel, no ELF executable or shared library file (including compiled Python C extension) found in the wheel archive
```

pyproject.toml
```diff
@@ -2,25 +2,23 @@
 name = "s2geometry"
 description = "Computational geometry and spatial indexing on the sphere"
 authors = [
-    { name = "Dan Larkin-York", email = "dan@arangodb.com" },
```
b4rdos, line 5: Did you use a formatter for this? Is the formatter tool / version recorded somewhere?

zacharyburnett: I used `taplo fmt` to format this

CMakeLists.txt
```diff
@@ -75,6 +75,10 @@ if (WITH_PYTHON)
     # Use Python3_ROOT_DIR to help find python3, if the correct location is not
     # being found by default.
     find_package(Python3 COMPONENTS Interpreter Development.Module)
+    find_package(PythonExtensions REQUIRED)
+    add_library(SwigBindings MODULE src/python)
+    python_extension_module(SwigBindings)
+    install(TARGETS SwigBindings LIBRARY DESTINATION s2geometry)
```
zacharyburnett: I'm not sure exactly what to put here; this just modifies the example from https://scikit-build.readthedocs.io/en/latest/usage.html#basic-usage

zacharyburnett:
unfortunately, using `scikit-build` still fails to provide the `s2geometry` module:
```
Run source .venv/bin/activate && python -c "import s2geometry"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
    import s2geometry
ModuleNotFoundError: No module named 's2geometry'
```

zacharyburnett:
building the Python 3.8 wheels for `macos-latest` and `macos-15-intel` fails with the following error:
https://github.com/google/s2geometry/actions/runs/24573839604/job/71853883913#step:3:612
```
error: aligned allocation function of type 'void *(std::size_t, std::align_val_t)' is only available on macOS 10.13 or newer
error: aligned deallocation function of type 'void (void *, std::align_val_t) noexcept' is only available on macOS 10.13 or newer
```

jmr:
> `error: aligned allocation function of type 'void *(std::size_t, std::align_val_t)' is only available on macOS 10.13 or newer`

Since the error is in absl, this is an absl bug, although I think other places in S2 will have this problem.

https://github.com/abseil/abseil-cpp#support

Support for 10.13 will be dropped in the summer, so I wouldn't worry too much about this one.

https://github.com/google/oss-policies-info/blob/main/foundational-cxx-support-matrix.md#macos

Why is this only a problem for this PR?

We can add
```
set(CMAKE_OSX_DEPLOYMENT_TARGET "10.13" CACHE STRING "Minimum macOS version")
```
before `project()`.

zacharyburnett:
Progress! the repair wheel step now fails on macOS with 
```
delocate.libsana.DelocationError: Failed to find any binary with the required architecture: 'arm64'
```
```
delocate.libsana.DelocationError: Failed to find any binary with the required architecture: 'x86_64'
```

zacharyburnett:
It looks like skipping wheel delocation for macOS lets `cibuildwheel` proceed, and the import tests pass with the built wheel; I'll test the uploaded artifacts on my arm64 machine when the workflow finishes

zacharyburnett:
testing these wheels locally on macOS ARM, the tests at `src/python/*_test.py` that imported `s2geometry` passed, but the ones that tried to import `s2geometry_pybind` failed (I assume because those need Bazel in order to be included in the wheel?). I think this PR is ready to merge for the SWIG bindings.

In order to publish to PyPI, @figroc just needs to configure the PyPI project https://pypi.org/project/s2geometry/ to add this repository (`https://github.com/google/s2geometry`) and build workflow (`build.yml`) to the list of Trusted Publishers. Then, on the next github release, the wheels will automatically upload to PyPI with the tagged version

zacharyburnett:
I don't know if C++ SWIG can make use of the "Limited API / Stable ABI" (https://docs.cython.org/en/latest/src/userguide/limited_api.html) but if so, that would make it so that only a single wheel would have to be built and published, instead of having to build a wheel for each Python version. But that would probably require changes in `src/python/` that are out of scope of this PR

figroc:
Done

> In order to publish to PyPI, @figroc just needs to configure the PyPI project https://pypi.org/project/s2geometry/ to add this repository (`https://github.com/google/s2geometry`) and build workflow (`build.yml`) to the list of Trusted Publishers.

pyproject.toml
```diff
@@ -60,3 +60,9 @@ namespaces = false
 # The same as the default.  See #450.
 # https://setuptools-scm.readthedocs.io/en/latest/config/#setuptools_scm._config.DEFAULT_TAG_REGEX
 tag_regex = "^(?:[\\w-]+-)?(?P<version>[vV]?\\d+(?:\\.\\d+){0,2}[^\\+]*)(?:\\+.*)?$"
+ 
+[tool.cibuildwheel]
+test-command = "python -c 'import s2geometry'"
```
jmr, line 65: can we actually run the tests here?

.github/workflows/build.yml
```diff
@@ -46,7 +46,7 @@ jobs:
       - run: mkdir build
       - run: >
           cmake
-          -DWITH_PYTHON=yes
+          -DWITH_PYTHON=OFF
```
jmr, line 49: why do you want to disable python here?

pyproject.toml
```diff
@@ -60,3 +60,9 @@ namespaces = false
...
+
+[tool.cibuildwheel.macos]
+repair-wheel-command = ""
```
jmr, line 68: comment what this is doing.  looks rather opaque.

CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
 cmake_minimum_required(VERSION 3.18)
+set(CMAKE_OSX_DEPLOYMENT_TARGET "10.13" CACHE STRING "Minimum macOS version")
```
jmr, line 2: Add a comment for why it's 13 and point to here:
https://github.com/google/oss-policies-info/blob/main/foundational-cxx-support-matrix.md#macos

setup.py
```diff
@@ -22,7 +22,9 @@
                                         '-DCALL_FROM_SETUP_PY:BOOL=ON',
                                         '-DBUILD_SHARED_LIBS:BOOL=OFF',
                                         '-DCMAKE_POSITION_INDEPENDENT_CODE=ON',
-                                        '-DWITH_PYTHON=ON'
+                                        '-DWITH_PYTHON=ON',
+                                        '-DBUILD_TESTS=OFF',
+                                        '-DFETCH_ABSEIL=ON',
```
jmr, line 27: Hmm. This seems like it could run us into problems with multiple absl verisons.

What shared libs do you end up with?

.github/workflows/build.yml
```diff
@@ -88,3 +88,51 @@ jobs:
...
+      - uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # v7.0.0
+        with:
+          name: dist-${{ runner.os }}-${{ runner.arch }}
+          path: ./dist/
+  python-publish:
+    if: (github.event_name == 'release') && (github.event.action == 'released')
+    needs: [python]
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
```
jmr, line 121: I need to learn more about these permissions. What's the alternative to putting this in a workflow?

jmr (changes requested):

jmr:
> I don't know if C++ SWIG can make use of the "Limited API / Stable ABI" (https://docs.cython.org/en/latest/src/userguide/limited_api.html) 

Yes, it does.  I'll see about adding this separately.

https://www.swig.org/Doc4.2/Python.html#Python_stable_abi

https://cmake.org/cmake/help/latest/module/FindPython.html#:~:text=Python%20embedding%20developments.-,Added%20in%20version%203.26%3A,-Development.SABIModule%3A%20search

.github/workflows/build.yml
zacharyburnett: this is to use Trusted Publishers with PyPI, see here: https://docs.github.com/en/actions/how-tos/secure-your-work/security-harden-deployments/oidc-in-pypi

from what I understand, after authenticating via OIDC, the workflow exchanges the token for a single-use PyPI publish token that it then uses to upload the wheels

.github/workflows/build.yml
zacharyburnett: good catch, I think this was from when I was testing changes to CMakeLists
```suggestion
          -DWITH_PYTHON=yes
```

jmr:
> Yes, it does. I'll see about adding this separately.

Added in 0b302d5

pyproject.toml
```diff
@@ -29,17 +29,17 @@ content-type = "text/plain"
 [project.urls]
 Source = "https://github.com/google/s2geometry"
 
-[project.optional-dependencies]
+[dependency-groups]
 test = [
     "pytest",
 ]
```
zacharyburnett, line 35: minor change: `project.optional-dependencies` are more for "features" visible to the user (they could do `pip install s2geometry[test]`); using `dependency-groups` denotes this as dependency for devs instead (install with `pip install --group test`)

setup.py
zacharyburnett: where would I find that? In the wheel itself? (but that wouldn't be shared would it?)

zacharyburnett: I thought this part of cmakelists was always fetching the same version of abseil so that it's statically included in the wheel: https://github.com/google/s2geometry/blob/0b302d5ab1cc3c13583b78ca7ba915e969f0c913/CMakeLists.txt#L66-L76

I apologize if that's wildly incorrect, I don't use cmake or compile C/++ code very much 😅

.github/workflows/build.yml
zacharyburnett: the alternative would be building the wheels locally, and then running `twine` or `uv publish` or something similar to push them to PyPI

src/s2geometry/include/absl/algorithm/algorithm.h
```diff
@@ -0,0 +1,64 @@
+// Copyright 2017 The Abseil Authors.
```
jmr: wait, you just added all of absl. what's going on?

.github/workflows/build.yml
```diff
@@ -94,3 +94,51 @@ jobs:
           --test_verbose_timeout_warnings \
           //...
       working-directory: src
+  python:
```
jmr, line 97: python should have timeout

src/s2geometry/include/absl/algorithm/algorithm.h
zacharyburnett: oh whoops, I need to add that to my gitignore, sorry!

pyproject.toml
jmr: what binary can't be found? what's the exact error message?

zacharyburnett: added the errors from https://github.com/google/s2geometry/pull/406#issuecomment-4281523497 to the comment in f2f51ea

pyproject.toml
```diff
@@ -60,3 +60,14 @@ namespaces = false
...
+ 
+[tool.cibuildwheel]
+# we only need to build with CPython 3.10 (because of the limited API / stable ABI)
+build = "cp310-*"
+test-groups = ["test"]
+test-requires = ["abi3audit"]
+test-command = "pytest {project}/src/python/s2geometry_test.py && abi3audit --strict --report {project}/dist/*.whl"
+
+[tool.cibuildwheel.macos]
+# skip wheel delocation (command can't find the binary)
```
jmr: Add:
```
  # delocate is not needed because BUILD_SHARED_LIBS is OFF in setup.py,
  # so all dependencies are statically linked into the extension module.
```

pyproject.toml
jmr: I don't know anything about this, but sounds better. Thanks!

.github/workflows/build.yml
jmr: The way you have it sounds fine.

setup.py
jmr: Since dynamic libs are disabled, this is fine.

jmr (approved):

zacharyburnett:
thank you for being so patient as I iterated over this. I'm excited to start using this in our applications!

jmr:
> thank you for being so patient as I iterated over this. 

Thanks for figuring this out!

> I'm excited to start using this in our applications!

I'll do a new release, hopefully later this week.

jmr:
https://github.com/google/s2geometry/releases/tag/v0.14.0

jmr:
failed: https://github.com/google/s2geometry/actions/runs/24833778811/job/72692695440

@figroc @zacharyburnett

figroc:
> failed: https://github.com/google/s2geometry/actions/runs/24833778811/job/72692695440
> 
> 
> 
> @figroc @zacharyburnett 

I've already add this repo (https://github.com/google/s2geometry) as alternate repository location in pypi project. Am I missing some configs? @zacharyburnett 

Or should I transfer the pypi project to some account? @jmr

zacharyburnett:
> > failed: https://github.com/google/s2geometry/actions/runs/24833778811/job/72692695440
> > @figroc @zacharyburnett
> 
> I've already add this repo (https://github.com/google/s2geometry) as alternate repository location in pypi project. Am I missing some configs? @zacharyburnett
> 
> Or should I transfer the pypi project to some account? @jmr

just to confirm, you added this github repo URL to the Trusted Publishers at https://pypi.org/manage/project/s2geometry/settings/publishing/ right?

it should have these values when filled out (`google`, `s2geometry`, `build.yml`, `pypi`): 
[image](https://pypi.org/manage/project/s2geometry/settings/publishing/)


Transferring the PyPI project also works! I don't know @jmr's PyPI username (if they have one), but if it's okay with @jmr my PyPI username is `zacharyburnett` and I can configure it

figroc:
> just to confirm, you added this github repo URL to the Trusted Publishers at https://pypi.org/manage/project/s2geometry/settings/publishing/ right?
> 
> it should have these values when filled out (`google`, `s2geometry`, `build.yml`, `pypi`): 
> 
> [image](https://pypi.org/manage/project/s2geometry/settings/publishing/)
> 
> Transferring the PyPI project also works! I don't know @jmr's PyPI username (if they have one), but if it's okay with @jmr my PyPI username is `zacharyburnett` and I can configure it

ok... I added to the wrong place. Now it should be good to go.

jmr:
seems to work now: https://pypi.org/project/s2geometry/

Issue #409, opened 2025-02-12
[CI] tests fail to build on macOS ARM64 runner


zacharyburnett:
The `cmake --build . --parallel` step takes 30 minutes and then is cancelled:
```
[ 66%] Building CXX object CMakeFiles/s2shapeutil_build_polygon_boundaries_test.dir/src/s2/s2shapeutil_build_polygon_boundaries_test.cc.o
[ 66%] Building CXX object CMakeFiles/value_lexicon_test.dir/src/s2/value_lexicon_test.cc.o
Error: The operation was canceled.
```
https://github.com/google/s2geometry/actions/runs/13096798651/job/36540005166#step:7:264

The same step takes ~3 minutes on macOS x64 and ~8 minutes on Ubuntu x64

zacharyburnett:
I'm not sure what is different about the ARM64 environment that specifically causes the build to be so slow. As far as I understand, the GitHub-provided ARM64 runner has 3 cores available (the x64 runner has 2). Perhaps the `cmake --parallel` is causing issues?

jmr:
Disabled by #412

PR #410, opened 2025-02-25
Fixed where `cc_configure_extension` comes from.

vtsao-openai:
Context: https://github.com/bazelbuild/bazel/issues/24426

Fixes `bazel mod deps`.

cc @romanoid-openai

### DIFF
--- a/src/MODULE.bazel
+++ b/src/MODULE.bazel
@@ -7,9 +7,15 @@ module(
     compatibility_level = 1,
 )
 
-cc_configure = use_extension("@bazel_tools//tools/cpp:cc_configure.bzl", "cc_configure_extension")
+cc_configure = use_extension("@rules_cc//cc:extensions.bzl", "cc_configure_extension")
 use_repo(cc_configure, "local_config_cc")
 
 bazel_dep(name = "boringssl", version = "0.0.0-20240126-22d349c")
-bazel_dep(name = "abseil-cpp", version = "20240722.0")
-bazel_dep(name = "googletest", version = "1.15.2")
+bazel_dep(name = "abseil-cpp", version = "20250127.0")
+bazel_dep(name = "googletest", version = "1.16.0")
+bazel_dep(name = "rules_cc", version = "0.1.1")
+
+# Forces usage of the patched re2 version that marks cc_configure_extension as a
+# dev dep as a workaround for https://github.com/bazelbuild/bazel/issues/24426.
+# TODO: remove once fixed upstream in re2, googletest, and abseil-cpp.
+bazel_dep(name = "re2", version = "2024-07-02.bcr.1")
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+bazel-*
+MODULE.bazel.lock


src/MODULE.bazel
```diff
@@ -7,9 +7,15 @@ module(
     compatibility_level = 1,
 )
 
-cc_configure = use_extension("@bazel_tools//tools/cpp:cc_configure.bzl", "cc_configure_extension")
+cc_configure = use_extension("@rules_cc//cc:extensions.bzl", "cc_configure_extension")
```
jmr, line 10: How long has `cc_configure_extension` been in `extensions.bzl`? Does this constrain the bazel version needed?

We should still work with bazel 7 LTS.

https://github.com/google/oss-policies-info/blob/main/foundational-cxx-support-matrix.md

src/MODULE.bazel
```diff
@@ -7,9 +7,15 @@ module(
...
 use_repo(cc_configure, "local_config_cc")
 
 bazel_dep(name = "boringssl", version = "0.0.0-20240126-22d349c")
-bazel_dep(name = "abseil-cpp", version = "20240722.0")
-bazel_dep(name = "googletest", version = "1.15.2")
+bazel_dep(name = "abseil-cpp", version = "20250127.0")
+bazel_dep(name = "googletest", version = "1.16.0")
+bazel_dep(name = "rules_cc", version = "0.1.1")
```
jmr, line 16: or is the bazel version irrelevant due to this line?

src/MODULE.bazel
vtsao-openai: Yeah since it's coming from `rules_cc` now I don't think it's tied to Bazel version anymore. I tested with Bazel 7 using:

`USE_BAZEL_VERSION=7.x bazel shutdown && USE_BAZEL_VERSION=7.x bazel mod deps`

and it worked.

src/MODULE.bazel
vtsao-openai: Yes I believe so, see my reply above.

vtsao-openai:
Not sure why the Mac cmake tests are failing, should be unrelated to Bazel changes?

jmr:
> Not sure why the Mac cmake tests are failing, should be unrelated to Bazel changes?

Yes, it's unrelated. See #409. I'll delete the Mac tests if we can't fix them soon.

If you added a bazel-based test, I wouldn't complain. :)

jmr (approved):

jmr:
Thanks!

src/MODULE.bazel
```diff
@@ -7,9 +7,15 @@ module(
...
+
+# Forces usage of the patched re2 version that marks cc_configure_extension as a
+# dev dep as a workaround for https://github.com/bazelbuild/bazel/issues/24426.
+# TODO: remove once fixed upstream in re2, googletest, and abseil-cpp.
```
jmr, line 20: @vtsao-openai https://github.com/google/re2/commit/c84a140 has been merged. What needs to happen now?

vtsao-openai: Ideally a new release of https://github.com/google/re2/releases so we can pull that new release into the BCR and then update `googletest` in the BCR to use the new version. I think it was googletest that was the problem in the dep graph: https://github.com/bazelbuild/bazel/issues/24426#issuecomment-2569718818

PR #418, opened 2025-05-04
Fix some comments in s2builderutil_snap_functions

rsned:


src/s2/s2builderutil_snap_functions.h
```diff
@@ -217,7 +217,7 @@ class IntLatLngSnapFunction : public S2Builder::SnapFunction {
   static int ExponentForMaxSnapRadius(S1Angle snap_radius);
 
   // For IntLatLng snapping, the minimum separation between vertices depends on
-  // exponent() and snap_radius().  It can vary between snap_radius()
+  // exponent() and snap_radius().  It can vary between 0.471 * snap_radius()
```
jmr, line 220: How about "It can vary between approximately 0.471 * snap_radius()"? Or "approximately sqrt(2)/3 * snap_radius()"?

jmr (changes requested):

src/s2/s2builderutil_snap_functions.h
rsned: updated.

jmr (approved):

Issue #411, opened 2025-02-28
S2 Master Fails to build with clang14


crroush:
There are numerous errors when building with g++12.3.0 in Ubuntu 22.04 off master [035272b63b313d2c79c7e8b9606f022b0570fe7b]

The following errors repeat a bunch of times.  Building using g++11.4 works fine.  
```
2-src/src/s2/base/port.h:35:3: error: no matching function for call to 'operator delete'
  ::operator delete(ptr, size);
  ^~~~~~~~~~~~~~~~~
/usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/new:180:13: note: candidate function not viable: no known conversion from 'size_t' (aka 'unsigned long') to 'void *' for 2nd argument; take the address of the argument with &
inline void operator delete  (void*, void*) _GLIBCXX_USE_NOEXCEPT { }
            ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/new:144:6: note: candidate function not viable: no known conversion from 'size_t' (aka 'unsigned long') to 'const std::nothrow_t' for 2nd argument
void operator delete(void*, const std::nothrow_t&) _GLIBCXX_USE_NOEXCEPT
     ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/new:153:6: note: candidate function not viable: no known conversion from 'size_t' (aka 'unsigned long') to 'std::align_val_t' for 2nd argument
void operator delete(void*, std::align_val_t)
     ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/new:130:6: note: candidate function not viable: requires 1 argument, but 2 were provided
void operator delete(void*) _GLIBCXX_USE_NOEXCEPT
     ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/new:155:6: note: candidate function not viable: requires 3 arguments, but 2 were provided
void operator delete(void*, std::align_val_t, const std::nothrow_t&)
     ^
```

jmr:
Try adding `#include <new>`.

jmr:
It works on godbolt: https://godbolt.org/z/Ta4dj9d8x

Can you run `cmake` with `-DCMAKE_VERBOSE_MAKEFILE=1`, then show the compilation command, and the `g++ --version` output from the `g++` that actually gets used?

Do you have a line like this new your `<new>`?

[...]

PR #421, opened 2025-05-17
CMakeLists: Make installation optional

idevtm:
Added install option in CMakeLists.txt to facilitate inclusion by submodule

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,7 +22,7 @@ endif()
 # this than assuming what absl used.  We default to c++17, but support
 # c++14 (like abseil).
 # Using CACHE allows the user to override the default.
-set(CMAKE_CXX_STANDARD 20 CACHE STRING "The C++ standard to build with")
+set(CMAKE_CXX_STANDARD 17 CACHE STRING "The C++ standard to build with")
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 # No compiler-specific extensions, i.e. -std=c++17, not -std=gnu++17.
 set(CMAKE_CXX_EXTENSIONS OFF)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,6 +42,12 @@ option(S2_USE_SYSTEM_INCLUDES
 add_feature_info(S2_USE_SYSTEM_INCLUDES S2_USE_SYSTEM_INCLUDES
     "Silence warnings in s2 headers by marking them as SYSTEM includes.")
 
+if(NOT CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR)
+  option(S2_ENABLE_INSTALL "Enable install rule" OFF)
+else()
+  option(S2_ENABLE_INSTALL "Enable install rule" ON)
+endif()
+
 feature_summary(WHAT ALL)
 
 # Don't output anything for LOG(INFO).
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,6 +42,8 @@ option(S2_USE_SYSTEM_INCLUDES
 add_feature_info(S2_USE_SYSTEM_INCLUDES S2_USE_SYSTEM_INCLUDES
     "Silence warnings in s2 headers by marking them as SYSTEM includes.")
 
+# when s2geometry is included as subproject (i.e. using add_subdirectory(s2geometry))
+# in the source tree of a project that uses it, install rules are disabled.
 if(NOT CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR)
   option(S2_ENABLE_INSTALL "Enable install rule" OFF)
 else()
@@ -277,7 +279,7 @@ set_target_properties(s2 PROPERTIES
     SOVERSION ${PROJECT_VERSION_MAJOR}
     VERSION ${PROJECT_VERSION})
 
-IF(S2_ENABLE_INSTALL)
+if(S2_ENABLE_INSTALL)
   # We don't need to install all headers, only those
   # transitively included by s2 headers we are exporting.
   install(FILES src/s2/_fp_contract_off.h
@@ -482,7 +484,7 @@ IF(S2_ENABLE_INSTALL)
           NAMESPACE s2::
           FILE s2Targets.cmake
           DESTINATION share/s2/)
-ENDIF() # S2_ENABLE_INSTALL
+endif() # S2_ENABLE_INSTALL
 
 if (BUILD_TESTS)
   if (NOT GOOGLETEST_ROOT)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,7 +22,7 @@ endif()
 # this than assuming what absl used.  We default to c++17, but support
 # c++14 (like abseil).
 # Using CACHE allows the user to override the default.
-set(CMAKE_CXX_STANDARD 17 CACHE STRING "The C++ standard to build with")
+set(CMAKE_CXX_STANDARD 20 CACHE STRING "The C++ standard to build with")
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 # No compiler-specific extensions, i.e. -std=c++17, not -std=gnu++17.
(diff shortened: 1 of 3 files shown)


src/s2/s2cell_id.cc
```diff
@@ -204,7 +204,7 @@ int S2CellId::GetCommonAncestorLevel(S2CellId other) const {
   // Compute the position of the most significant bit, and then map the bit
   // position as follows:
   // {0} -> 30, {1,2} -> 29, {3,4} -> 28, ... , {59,60} -> 0, {61,62,63} -> -1.
-  return max(61 - absl::bit_width(bits), -1) >> 1;
+  return max(61 - absl::bit_width(bits), -1ul) >> 1;
```
jmr, line 207: Why is this changing?

src/s2/base/port.h
```diff
@@ -32,11 +32,19 @@
 // C++14 sized deallocation
 namespace base {
 inline void sized_delete(void *ptr, size_t size) {
+#if defined(__cpp_sized_deallocation)
   ::operator delete(ptr, size);
```
jmr, line 36: This is C++14, so why wouldn't it be defined?  

It appears there's a bug in old clang versions that doesn't enable sized delete when it should though.

https://github.com/google/s2geometry/issues/411

CMakeLists.txt
```diff
@@ -271,210 +271,212 @@ set_target_properties(s2 PROPERTIES
...
-install(FILES src/s2/util/units/length-units.h
-              src/s2/util/units/physical-units.h
-        DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/s2/util/units")
-
-if (GOOGLETEST_ROOT)
-  set(S2_TARGETS s2 s2testing)
-else()
-  set(S2_TARGETS s2)
-endif()
+IF(S2_ENABLE_INSTALL)
```
jmr, line 274: I don't understand how this is different from not typing `make install`. Could you explain more about why you want this?

jmr:
This PR does 3 things. It would be better to split it up, but maybe some parts are not necessary anyway.

https://github.com/google/s2geometry/issues/411

jmr (changes requested):

src/s2/s2cell_id.cc
idevtm:
```
error: no matching function for call to 'max'
[build]   return max(61 - absl::bit_width(bits), -1) >> 1;
[build]          ^~~
[build] /usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/bits/algorithmfwd.h:407:5: note: candidate template ignored: deduced conflicting types for parameter '_Tp' ('unsigned long' vs. 'int')
```

I use Clang15 on Debian 12 and it looks like it's not able to resolve this correctly. however a static_cast<int> would be better since `-1ul` is not really correct.

src/s2/base/port.h
idevtm: indeed Clang15.0.6 complains about it

CMakeLists.txt
idevtm: I have added S2 (and absl) as submodules in ext/s2geometry and attempt to use them by including this .cmake; then I can link `ext-s2geometry` in various project targets after `include(ext/s2geometry.cmake)`.
```
if(TARGET ext-s2geometry)
    return()
endif()

# Abseil first
set(ABSL_SRC_DIR ${CMAKE_CURRENT_LIST_DIR}/absl)
set(ABSL_BIN_DIR ${CMAKE_BINARY_DIR}/_deps/absl-build)
set(ABSL_ENABLE_INSTALL ON CACHE BOOL "" FORCE)
add_subdirectory(${ABSL_SRC_DIR} ${ABSL_BIN_DIR})

# Add the S2 library
set(S2GEOM_SRC_DIR ${CMAKE_CURRENT_LIST_DIR}/s2geometry)
set(S2GEOM_BIN_DIR ${CMAKE_BINARY_DIR}/_deps/s2geometry-build)
set(S2_ENABLE_INSTALL OFF CACHE BOOL "" FORCE)
set(BUILD_TESTS OFF CACHE BOOL "" FORCE)
add_subdirectory(${S2GEOM_SRC_DIR} ${S2GEOM_BIN_DIR})

# Create an alias target for cleaner usage
add_library(ext-s2geometry ALIAS s2)
```

however without `S2_ENABLE_INSTALL OFF` installation is attempted by default and it's not working, cmake errors pasted below:
```
[cmake] CMake Error in ext/s2geometry/CMakeLists.txt:
[cmake]   Target "s2" INTERFACE_INCLUDE_DIRECTORIES property contains path:
[cmake] 
[cmake]     ".../ext/s2geometry/src"
[cmake] 
[cmake]   which is prefixed in the source directory.
```

jmr: Can you do something like abseil-cpp does?

https://github.com/abseil/abseil-cpp/blob/41a1f434d40bc6089acaf026b00c5f8f917e8706/CMakeLists.txt#L38

src/s2/s2cell_id.cc
jmr: This implies that `absl::bit_width` is returning `unsigned long`, but it should be returning `int`.  Are you in the code path using `std::bit_width` or the `absl::` one?

https://github.com/abseil/abseil-cpp/blob/41a1f434d40bc6089acaf026b00c5f8f917e8706/absl/numeric/bits.h#L144

Please split this to another PR since it's not related to cmake header installation logic.

src/s2/base/port.h
jmr: Revert this, I'll fix #411 separately.

jmr (changes requested):

CMakeLists.txt
```diff
@@ -42,6 +42,12 @@ option(S2_USE_SYSTEM_INCLUDES
 add_feature_info(S2_USE_SYSTEM_INCLUDES S2_USE_SYSTEM_INCLUDES
     "Silence warnings in s2 headers by marking them as SYSTEM includes.")
 
+if(NOT CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR)
```
jmr, line 45: Please add a comment (abseil-cpp had one), it is not at all obvious what this is doing here.

CMakeLists.txt
```diff
@@ -271,210 +277,212 @@ set_target_properties(s2 PROPERTIES
...
-install(FILES src/s2/util/units/length-units.h
-              src/s2/util/units/physical-units.h
-        DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/s2/util/units")
-
-if (GOOGLETEST_ROOT)
-  set(S2_TARGETS s2 s2testing)
-else()
-  set(S2_TARGETS s2)
-endif()
+IF(S2_ENABLE_INSTALL)
```
jmr, line 280: the rest of the files uses lower case `if`/`endif`.  Be consistent.

src/s2/s2cell_id.cc
idevtm: This it taking the path with `using std::bit_width;`. 
I'm splitting to another PR.

jmr (approved):

src/s2/s2cell_id.cc
jmr: libstdc++ used to have the wrong return type.

https://github.com/gcc-mirror/gcc/commit/4581328022615c2cb732cfb404151c4b059d26e3

I wonder if `abseil-cpp` should avoid `std::bit_width` or add a wrapper with a cast for buggy libstdc++ versions.  Can you file a bug with abseil-cpp?

jmr: https://github.com/abseil/abseil-cpp/issues/1890

idevtm: `-1ul` may have the unintended wrap around value of `0xFFFFFFFF...`; we may want to use `static_cast<int>(...)` on the left operand instead

jmr: Yes, it should be `static_cast<int>(absl::bit_width(...))` with a comment explaining the workaround.

PR #433, opened 2025-05-24
Try to fix CMakeLists.txt

MBkkt:
I'm not CMake expert, but with current code and WITH_PYTHON not touched  I have following error:
```
CMake Error at third_party/s2geometry/CMakeLists.txt:658 (if):
  if given arguments:

    "AND" "TRUE"

  Unknown arguments specified
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -655,7 +655,7 @@ if (BUILD_EXAMPLES AND TARGET s2testing)
   add_subdirectory("doc/examples" examples)
 endif()
 
-if (${SWIG_FOUND} AND ${Python3_FOUND})
+if (SWIG_FOUND AND Python3_FOUND)
   add_subdirectory("src/python" python)
 endif()
 


CMakeLists.txt
```diff
@@ -655,7 +655,7 @@ if (BUILD_EXAMPLES AND TARGET s2testing)
   add_subdirectory("doc/examples" examples)
 endif()
 
-if (${SWIG_FOUND} AND ${Python3_FOUND})
+if (SWIG_FOUND AND Python3_FOUND)
```
jmr, line 658: Odd since this hasn't changed in 5 years.

https://github.com/google/s2geometry/blame/473a989af642959b1a52bdb675445ff1a9dd9c09/CMakeLists.txt#L658

What's the exact command that fails for you and what cmake version?

MBkkt: It can became an issue since this https://github.com/google/s2geometry/commit/efb124d8eaf3433323d3e877dedd5e94a63339a3

MBkkt: Maybe you can try following patch in github CI?
https://github.com/google/s2geometry/compare/master...MBkkt:s2geometry:patch-11

jmr: The easiest way to test it is to make a PR.

MBkkt: I think I find main difference in `if (condition)` https://cmake.org/cmake/help/latest/command/if.html#variable-expansion but still don't understand why `if (${undefined_var} AND ${defined_as_true_var})` doesn't work

PR #436, opened 2025-06-11
Fix warnings

MBkkt:
Produced by this `ABSL_UNLOCK_FUNCTION(lock_)` for thread analysis noone called Lock, so it thinks unlock called without lock

```cpp
#ifndef SWIG
  // Documented in the .cc file.
  void UnlockAndSignal() ABSL_UNLOCK_FUNCTION(lock_)
      ABSL_UNLOCK_FUNCTION(update_state_->wait_mutex);
#endif
```

### DIFF
--- a/src/s2/base/spinlock.h
+++ b/src/s2/base/spinlock.h
@@ -16,28 +16,29 @@
 #ifndef S2_BASE_SPINLOCK_H_
 #define S2_BASE_SPINLOCK_H_
 
+#include <absl/base/thread_annotations.h>
+
 #include <atomic>
 
-
-class SpinLock {
+class ABSL_LOCKABLE SpinLock {
  public:
   SpinLock() = default;
   ~SpinLock() = default;
   SpinLock(SpinLock const&) = delete;
   SpinLock& operator=(SpinLock const&) = delete;
 
-  inline void Lock() {
+  inline void Lock() ABSL_EXCLUSIVE_LOCK_FUNCTION() {
     while (locked_.exchange(true, std::memory_order_acquire)) {
       // Spin.
       continue;
     }
   }
 
-  inline void Unlock() {
+  inline void Unlock() ABSL_UNLOCK_FUNCTION() {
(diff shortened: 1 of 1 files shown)


jmr (approved):

src/s2/base/spinlock.h
```diff
@@ -16,39 +16,43 @@
...
     }
   }
 
-  inline void Unlock() {
+  inline void Unlock() ABSL_UNLOCK_FUNCTION() {
     locked_.store(false, std::memory_order_release);
   }
 
-  inline bool IsHeld() const {
+  inline bool IsHeld() const ABSL_ASSERT_EXCLUSIVE_LOCK() {
```
jmr, line 41: I didn't look at this closely enough.

`ABSL_ASSERT_EXCLUSIVE_LOCK` is for a function that `assert()`s / terminates.

https://clang.llvm.org/docs/ThreadSafetyAnalysis.html#assert-capability-and-assert-shared-capability

Can you remove `ABSL_ASSERT_EXCLUSIVE_LOCK()` for now or make an `AssertHeld()` function and use that?

https://github.com/abseil/abseil-cpp/blob/9ac131cf7da4de8b19b4c956a0c13edba74d92d9/absl/base/internal/spinlock.h#L129

jmr: `AssertHeld()` asserts in `opt` mode as well as `dbg`, so just remove `ABSL_ASSERT_EXCLUSIVE_LOCK`.

MBkkt: https://github.com/google/s2geometry/pull/437

Made PR to remove incorrect annotation.

MBkkt: About assertheld I'm not sure, because usecase in s2 different:
`ABSL_DCHECK(cells_lock_.IsHeld());`

MBkkt: Anyway it compiles without any declaration for `IsHeld()` so it's probably ok to keep it like this for now?

PR #440, opened 2025-07-10

jmr:
* Require C++17
* Require abseil-cpp LTS 20250512
* Fix S2ShapeNestingQuery when two chains share a vertex
* Fix S2Polygon bug when intersecting polyline
* Add S2ClosestEdgeQuery::VisitClosestShapes
* Delete deprecated S2Error::text(); use message() instead
* Ensure _fp_contract_off.h is included everywhere
* Add thread annotations to SpinLock
* Convert many const vector<>&s to Span<>s
* Add noexcept to move operators
* Make some comparisons stable for sorting
* Guard against overflow in Decode functions


jmr:
@smcallis

PR #460, opened 2025-10-04
Add config for ASAN builds with bazel.

smcallis:
Enables ASAN builds with e.g. bazel -c opt --config=asan <targets>


src/.bazelrc
```diff
@@ -1,3 +1,12 @@
...
-common --cxxopt=-std=c++20
\ No newline at end of file
+common --cxxopt=-std=c++20
+
+build:asan --strip=never
+build:asan --copt -fsanitize=address
+build:asan --copt -DADDRESS_SANITIZER
+build:asan --copt -D_GLIBCXX_SANITIZE_VECTOR
+build:asan --copt -O1
+build:asan --copt -g
```
jmr: A lot of these seem orthogonal and part of `-c dbg`.

src/.bazelrc
```diff
@@ -1,3 +1,12 @@
 # Enable Bzlmod for every Bazel command
 common --enable_bzlmod
-common --cxxopt=-std=c++20
\ No newline at end of file
+common --cxxopt=-std=c++20
+
+build:asan --strip=never
+build:asan --copt -fsanitize=address
+build:asan --copt -DADDRESS_SANITIZER
+build:asan --copt -D_GLIBCXX_SANITIZE_VECTOR
```
jmr: In a few years, it will be a mystery why these are here. Document.
```
# Avoid container overflow false positives.
# https://github.com/google/sanitizers/wiki/AddressSanitizerContainerOverflow#false-positives
```

src/.bazelrc
```diff
@@ -1,3 +1,12 @@
 # Enable Bzlmod for every Bazel command
 common --enable_bzlmod
-common --cxxopt=-std=c++20
\ No newline at end of file
+common --cxxopt=-std=c++20
+
+build:asan --strip=never
+build:asan --copt -fsanitize=address
+build:asan --copt -DADDRESS_SANITIZER
```
jmr: Where is this used?

src/.bazelrc
```diff
@@ -1,3 +1,12 @@
 # Enable Bzlmod for every Bazel command
 common --enable_bzlmod
-common --cxxopt=-std=c++20
\ No newline at end of file
+common --cxxopt=-std=c++20
+
+build:asan --strip=never
+build:asan --copt -fsanitize=address
```
jmr: use `--arg=value` consistently

src/.bazelrc
smcallis: You can theoretically use this with -c opt or -c fastbuild too, it's probably wise to override them because if you have e.g. -O3 on and the compiler elides problematic reads/writes then you'll get false negatives.

src/.bazelrc
smcallis: done

src/.bazelrc
smcallis: I added a comment you use it with --config=asan when running bazel.

src/.bazelrc
smcallis: Done.

Issue #467, opened 2025-10-31
Optimize S2Cell::GetDistance


jmr:
@ericveach Do you remember what you had in mind with
> This could be optimized to be at least 5x faster by pruning the set of possible closest vertex/edge pairs using the faces and (u,v) ranges of both cells.

?

https://github.com/google/s2geometry/blob/a538e026bfd235092ab1a848bf5a2a517ecb2e11/src/s2/s2cell.cc#L508-L526

jmr:
> it's also not too hard to test for the condition where the minimum distance is attained between two vertices rather than a vertex and an edge.

Would this be testing one pair of vertices or multiple pairs? If the former, that sounds very good.

PR #479, opened 2025-11-24
S2Cell::GetDistance: Optimize same-face case

jmr:
Use the uv coordinates to prune the vertex-edge distance computations needed.

When cell A is above B, we don't need to compare the top vertices/edge of A with the bottom vertices/edge of B.

Note, however, that when A is both above and left of B, we cannot just compute the lower-right/upper-left vertex distance due to the projection.

It is possible that more comparisons can be omitted; I'm not sure.

Error is within 1e-15 radians of previous results.

GetDistance shows a 2x speedup for same-face cells.

### DIFF
--- a/src/s2/s2cell.cc
+++ b/src/s2/s2cell.cc
@@ -639,21 +639,14 @@ bool S2Cell::IsDistanceLess(const S2Cell& target, S1ChordAngle limit) const {
     const S2Point va2 = GetVertex(ai + 1);
     const S2Point vb1 = target.GetVertex(bi);
     const S2Point vb2 = target.GetVertex(bi + 1);
-    if (S2::UpdateMinDistance(va1, vb1, vb2, &min_dist) && min_dist < limit) {
-      return true;
-    }
-    if (S2::UpdateMinDistance(va2, vb1, vb2, &min_dist) && min_dist < limit) {
-      return true;
-    }
-    if (S2::UpdateMinInteriorDistance(vb1, va1, va2, &min_dist) &&
-        min_dist < limit) {
-      return true;
-    }
-    if (S2::UpdateMinInteriorDistance(vb2, va1, va2, &min_dist) &&
-        min_dist < limit) {
-      return true;
-    }
-    return false;
+    return (S2::UpdateMinDistance(va1, vb1, vb2, &min_dist) &&
+            min_dist < limit) ||
+           (S2::UpdateMinDistance(va2, vb1, vb2, &min_dist) &&
+            min_dist < limit) ||
+           (S2::UpdateMinInteriorDistance(vb1, va1, va2, &min_dist) &&
+            min_dist < limit) ||
+           (S2::UpdateMinInteriorDistance(vb2, va1, va2, &min_dist) &&
+            min_dist < limit);
   }
 
   // Otherwise, the minimum distance always occurs between a vertex of one


jmr:
Thanks, Eric! https://github.com/google/s2geometry/pull/479#issuecomment-3586683570 works for me after I swapped `kTopEdge` and `kBottomEdge` and used `(_ + 1) & 3` instead of `_ + 1`.  I tried something similar after I noticed [the test failures were far in one direction and not the other](https://github.com/google/s2geometry/issues/467#issuecomment-3581471252), but I must have made a mistake when I tried it.

jmr:
> Although I can't build the actual library right now I did try pasting some version of the code I suggested into compiler explorer, and when built under Clang the FindFurthestEdge portion is branchless. Given this and the fact that only 4 edges are tested, how much faster is it now?

`UpdateMinDistance` is a 3.8x speedup, and `GetVertex` is 1.5x.  Based on operation count, 8x and 2x would be ideal.

Before it was 80% `UpdateMinDistance` and 15% `GetVertex`, after it's 60%30%.

This gives an overall 3x speedup.  We can take the win and continue to optimize.  The largest chunk of the remaining time is in vector `Normalize` and `Norm2`.

> Unfortunately GCC doesn't seem willing to generate branchless code, but there's not much we can do about that.

I can get gcc to generate branchless code by giving it a branch probability somewhere between 0.21 and 0.33.

https://godbolt.org/z/rhqGcrqE9

It would be interesting to see the actual branch probability and what FDO did.

ericveach:
Thanks for the comprehensive analysis!  It looks like the performance analysis tools available have improved a lot in the past few years.

Just to clarify, are the results you mentioned for random cells, or for cells on the same face, or for cells that are small relative to their separation distance?  I suspect that the last case may be the most important in practice, e.g. where the separation distance may be up to a few hundred or thousands of km but the cell sizes are, say, only 1% or 10% of that distance.  This is the situation you would often have when measuring distances between coverings of real-world geometry, for example.

ericveach:
Maybe your existing change is big enough, but if you wanted to try adding the vertex-only case, here is a stab at it:

```
  if (face_ == target.face_) {
    // In certain cases the distance between two cells is attained between a     
    // pair of vertices.  This makes the distance very cheap to compute and so   
    // it's worth detecting the easy cases where this happens.  Recall that      
    // all cells except at level 0 are slightly diamond-shaped, i.e. one         
    // diagonal is slightly longer than the other and the corresponding cell     
    // corners are either right-angled or acute.  Define a diagonal to have      
    // positive slope if u and v both increase along it, and negative slopee     
    // otherwise.  Furthermore define two cells to be separated along a          
    // positive diagonal if one has strictly larger u- and v-values than the     
    // other, and along a negative diagonal if one cell has strictly smaller     
    // u-values and strictly larger v-values or vice versa.  Then if two cells   
    // A and B are separated along a positive diagonal and the long diagonals    
    // of A and B both have positive slope, then the minimum distance occurs     
    // between two vertices; namely, the closer endpoints of their long          
    // diagonals.  The same is true if the two cells are separated along a       
    // negative diagonal and both long diagonals have negative slope.  One of    
    // these two situations is expected to occur almost 50% of the time when     
    // the cells are on the same face and are small relative to their            
    // separation distance.                                                      
    //                                                                           
    // For the purpose of determining whether cells are separated along a        
    // diagonal, as described above, we assign constants to the four edges of    
    // A as follows: L=1, T=3, R=3, B=9.  "sep_dirs" is the sum of these         
    // values for the edges of A that separate A from B.  This yields the        
    // following possible sums: TL=4, TR=6, BL=10, BR=12, L=1, T=3, R=3, B=9,    
    // none=0.  This lets us test the following conditions cheaply, given that   
    // we test them in the following order:                                      
    //                                                                           
    //  sep_dirs == 0 : the two cells intersect                                  
    //  !(sep_dirs & 1) : cells are separated along a diagonal (TL, TR, BL, BR)  
    //  (sep_dirs & 2) : cells are separated along a positive diagonal (TL, BR)  
    //  (sep_dirs < 8) : cell A  cell B (TL, TR)                                 

    int sep_dirs = 0;
    R2Rect a_uv = a.GetBoundUV();
    R2Rect b_uv = b.GetBoundUV();
    if (a_uv[0][0] > b_uv[0][1]) sep_dirs += 1;  // left side of A               
    if (a_uv[0][1] < b_uv[0][0]) sep_dirs += 3;  // right                        
    if (a_uv[1][0] > b_uv[1][1]) sep_dirs += 9;  // bottom                       
    if (a_uv[1][1] < b_uv[1][0]) sep_dirs += 3;  // top                          
    if (sep_dirs == 0) {
      // A and B intersect (this includes edge and vertex intersections).        
      return S1ChordAngle::Zero();
    }
    // Otherwise if the two cells are separated along a diagonal, check if the   
    // cells also have their long diagonals in that direction.                   
    bool separated = !(sep_dirs & 1);  // TL, TR, BL, BR                         
    if (separated) {
      bool sep_positive = (sep_dirs & 2) != 0;  // TR or BL                      
      bool a_positive = (a_uv[0][0] < 0) != (a_uv[1][0] < 0);
      bool b_positive = (b_uv[0][0] < 0) != (b_uv[1][0] < 0);
      // The use of & rather than && below encourages branchless compilation.    
      if (a_positive == sep_positive & b_positive == sep_positive) {
        // Compute the vertex of A that is closest to B, without branches.       
        // Vertices are numbered as follows: BL=0, BR=1, TR=2, TL=3.             
        // a_positive implies that the closest vertex to B is TR or BL.          
        // a_top_edge implies that the closest vertex to B is TR or TL.          
        bool a_top_edge = sep_dirs < 8;  // TR or TL                             
        int i = (a_positive ? 0 : 1) + (a_top_edge ? 2 : 0);
        return S1ChordAngle(GetVertex(i), target_.GetVertex(i ^ 2));
      }
    }
    // Otherwise carry on with the existing code, except that the (ai < 0) case is no longer needed.
  }

```

Again, no guarantees that this will work or even compile.  But at least in the case of cells that are small relative to their separation distance, and where the separation distance is small relative to the Earth's radius, it should yield a substantial speedup.

src/s2/s2cell.cc
```diff
@@ -497,27 +497,73 @@ S1ChordAngle S2Cell::GetMaxDistance(const S2Point& a, const S2Point& b) const {
   return S1ChordAngle::Straight() - GetDistance(-a, -b);
 }
 
+namespace {
+// Finds the index `ai` of the edge of A that is furthest away from the
+// opposite edge of B in (u,v)-space and B is in that direction from A.
```
ericveach: "such that" or "where" would be better than "and" here.

src/s2/s2cell.cc
```diff
@@ -497,27 +497,73 @@ S1ChordAngle S2Cell::GetMaxDistance(const S2Point& a, const S2Point& b) const {
...
+    // edge `ai` and the opposite edge (`ai ^ 2`) of B, or symmetrically, an
+    // endpoint of the opposite edge of B and the edge `ai`.
+    int bi = ai ^ 2;
+    const S2Point va1 = GetVertex(ai);
+    const S2Point va2 = GetVertex(ai + 1);
+    const S2Point vb1 = target.GetVertex(bi);
+    const S2Point vb2 = target.GetVertex(bi + 1);
+    S2::UpdateMinDistance(va1, vb1, vb2, &min_dist);
+    S2::UpdateMinDistance(va2, vb1, vb2, &min_dist);
+    S2::UpdateMinDistance(vb1, va1, va2, &min_dist);
```
ericveach: Technically you could use UpdateMinInteriorDistance() for the last two, but I doubt it will make a noticeable difference since UpdateMinDistance() calculates the distance to the edge endpoints anyway; it essentially would only save two "min" calculations.

src/s2/s2cell_test.cc
```diff
@@ -539,6 +540,102 @@ static S1ChordAngle GetMaxDistanceToPointBruteForce(const S2Cell& cell,
   return max_distance;
 }
 
+// Previous implementation of `S2Cell::GetDistance(const S2Cell&)`.
+static S1ChordAngle GetDistanceToCellBruteForce(const S2Cell& cell1,
+                                                const S2Cell& cell2) {
```
ericveach: With the va/vb naming below, it might be clearer to call the arguments "a" and "b".

src/s2/s2cell_test.cc
```diff
@@ -539,6 +540,102 @@ static S1ChordAngle GetMaxDistanceToPointBruteForce(const S2Cell& cell,
...
+                     << ((j + 1) & 3) << " min_dist: " << min_dist.degrees();
+      }
+    }
+  }
+  return min_dist;
+}
+
+TEST(S2Cell, GetDistanceToCell) {
+  absl::BitGen bitgen(S2Testing::MakeTaggedSeedSeq(
+      "GET_DISTANCE_TO_CELL", absl::LogInfoStreamer(__FILE__, __LINE__).stream()));
```
ericveach, line 579: That's quite a mouthful.

src/s2/s2cell_test.cc
```diff
@@ -539,6 +540,102 @@ static S1ChordAngle GetMaxDistanceToPointBruteForce(const S2Cell& cell,
...
+        << "cell1: " << cell1.id() << " cell2: " << cell2.id()
+        << " cell1.uv: " << cell1.GetBoundUV()
+        << " cell2.uv: " << cell2.GetBoundUV();
+  }
+}
+
+TEST(S2Cell, GetDistanceToCellHighErrorExample) {
+  // This is a test case extracted from `GetDistanceToCell`; it achieved
+  // the maximum error over 100M iterations, so is useful to understand
+  // the errors and as a shortcut for testing.
```
ericveach: This is probably a useful test case, but it's not really measuring error, just the difference in error between two different calculation methods.  S2::GetUpdateMinDistanceMaxError() returns a provable upper bound on the actual error.

ericveach (approved):

src/s2/s2cell.cc
jmr: Done.

src/s2/s2cell_test.cc
jmr: Done.

src/s2/s2cell_test.cc
jmr: In google3, it's just `absl::LogInfoStreamer().stream()`. When we can depend on C++20, we can probably do that here, too.

src/s2/s2cell_test.cc
jmr: Ok, I renamed it to `GetDistanceToCellHighDifferenceExample`.

Is there any way to usefully use `GetUpdateMinDistanceMaxError()` in a test?

src/s2/s2cell.cc
jmr: Actually, this does make quite a nice difference:

```
name                                                   CYCLES/op   CYCLES/op   vs base              
BM_GetDistanceToCell                                    755.4 ± 0%   680.2 ± 1%  -9.95% (p=0.002 n=6)
BM_GetDistanceToCellSameFace                            225.4 ± 0%   211.7 ± 0%  -6.07% (p=0.002 n=6)
BM_DistanceCompare<&S2Cell::IsDistanceLess>/0           8.066 ± 0%   8.117 ± 1%       ~ (p=0.182 n=6)
BM_DistanceCompare<&S2Cell::IsDistanceLess>/10          216.8 ± 0%   197.7 ± 1%  -8.79% (p=0.002 n=6)
BM_DistanceCompare<&S2Cell::IsDistanceLess>/50          185.4 ± 2%   175.1 ± 1%  -5.58% (p=0.002 n=6)
BM_DistanceCompare<&S2Cell::IsDistanceLess>/90          124.0 ± 1%   122.3 ± 1%  -1.32% (p=0.002 n=6)
BM_DistanceCompare<&S2Cell::IsDistanceLess>/100         112.5 ± 0%   112.4 ± 0%       ~ (p=1.000 n=6)
BM_DistanceCompare<&S2Cell::IsDistanceLessOrEqual>/0    225.5 ± 2%   206.7 ± 2%  -8.34% (p=0.002 n=6)
BM_DistanceCompare<&S2Cell::IsDistanceLessOrEqual>/10   225.3 ± 1%   206.7 ± 2%  -8.25% (p=0.002 n=6)
BM_DistanceCompare<&S2Cell::IsDistanceLessOrEqual>/50   224.1 ± 1%   204.9 ± 1%  -8.54% (p=0.002 n=6)
BM_DistanceCompare<&S2Cell::IsDistanceLessOrEqual>/90   224.2 ± 1%   206.5 ± 2%  -7.86% (p=0.002 n=6)
BM_DistanceCompare<&S2Cell::IsDistanceLessOrEqual>/100  224.1 ± 1%   204.4 ± 1%  -8.76% (p=0.002 n=6)
geomean                                                166.1        155.9       -6.14%
```

It looks like this can also be used in the fallback, all pairs case.

jmr: Please take another look since I used UpdateMinInteriorDistance in additional places.

src/s2/s2cell.cc
```diff
@@ -568,8 +616,68 @@ S1ChordAngle S2Cell::GetMaxDistance(const S2Cell& target) const {
 }
 
 bool S2Cell::IsDistanceLess(const S2Cell& target, S1ChordAngle limit) const {
-  // TODO: b/284470618 - Use more efficient implementation
-  return GetDistance(target) < limit;
+  // Implementation of this function is similar to `GetDistance(target)`, but
```
jmr, line 619: I also added this function

src/s2/s2cell.cc
```diff
@@ -614,8 +616,68 @@ S1ChordAngle S2Cell::GetMaxDistance(const S2Cell& target) const {
...
+  }
+  for (int i = 0; i < 4; ++i) {
+    for (int j = 0; j < 4; ++j) {
+      if (S2::UpdateMinDistance(va[i], vb[j], vb[(j + 1) & 3], &min_dist) &&
+          min_dist < limit) {
+        return true;
+      }
+      if (S2::UpdateMinInteriorDistance(vb[i], va[j], va[(j + 1) & 3],
+                                        &min_dist) &&
+          min_dist < limit) {
```
ericveach, line 668: The line-splitting is pretty awful, but I presume this is the current clang-format standard.  A shorter name for "min_dist" would help (such as "dist") but I'm not sure it's worth it.  (Although technically the adjective "min" is superfluous when speaking of the distance between two objects; it refers to the minimum distance by default unless you add some other adjective such as "maximum" or "Hausdorff".)  A shorter name would also let you do the following above, clang-format willing:

```
    return (S2::UpdateMinDistance(va1, vb1, vb2, &dist) && dist < limit)
      || (S2::UpdateMinDistance(va2, vb1, vb2, &dist) && dist < limit)
      || (S2::UpdateMinInteriorDistance(vb1, va1, va2, &dist) && dist < limit)
      || (S2::UpdateMinInteriorDistance(vb2, va1, va2, &dist) && dist < limit);

```

src/s2/s2cell_test.cc
ericveach: If you grep for this function (or use whatever fancy IDE tools you have available these days) you'll find a number of uses in tests.  It's useful when you want to run a conservative query that is guaranteed to find all objects whose *true* distance from a reference object is less than a given threshold, e.g. because you then want to use a function such as s2pred::CompareEdgeDistance to check whether the exact distance is above or below a threshold.  This sort of thing is required by classes like S2Builder in order to implement certain algorithms so that they can make absolute guarantees about their results.

There are already tests elsewhere to check that the bounds returned by GetUpdateMinDistanceMaxError() are accurate, so there is no need to repeat them here.

src/s2/s2cell.cc
jmr: Writing it as one expression with the current names gives:
```c++
    return (S2::UpdateMinDistance(va1, vb1, vb2, &min_dist) &&
            min_dist < limit) ||
           (S2::UpdateMinDistance(va2, vb1, vb2, &min_dist) &&
            min_dist < limit) ||      
           (S2::UpdateMinInteriorDistance(vb1, va1, va2, &min_dist) &&
            min_dist < limit) ||
           (S2::UpdateMinInteriorDistance(vb2, va1, va2, &min_dist) &&
            min_dist < limit);
```

`min_dist` and `max_dist` are used consistently in this file, I assume to catch things like `UpdateMaxDistance(..., &min_dist)`.  Such a mistake seems unlikely, and there are tests.

If we consistently replace with just `d`, this gets formatted as:
```c++
    return (S2::UpdateMinDistance(va1, vb1, vb2, &d) && d < limit) ||
           (S2::UpdateMinDistance(va2, vb1, vb2, &d) && d < limit) ||
           (S2::UpdateMinInteriorDistance(vb1, va1, va2, &d) && d < limit) ||
           (S2::UpdateMinInteriorDistance(vb2, va1, va2, &d) && d < limit);
```
I think the scope is small enough for `d`, but the google3 reviewers might not.

We aren't beholden to clang-format and there are hundreds of options to tweak it.  The operators are very consistently at the end of the line, not the beginning, but this may be a hobgoblin of little minds thing.

ericveach: Any way you'd like to handle this is fine with me, including leaving it the way it was - thanks!

jmr: I changed it to one expression, but with the `min_dist` name and formatting for now.

src/s2/s2cell_test.cc
jmr: Ok. I thought you wanted me to add a test using `GetUpdateMinDistanceMaxError`.

PR #480, opened 2025-11-27
Add s2geometry.io website source from github.com/manshreck/s2geometry.github.io

ivan-gerasimov:
Hi team,

This pull request adds the source code for the s2geometry.io website. The goal is to migrate the site to a Google-controlled repository. This will allow us to fix the HTTPS certificate issue and resolve the "Not Secure" browser warning.

### Directory Choice: `/docs`

The website source files have been placed in the `/docs` directory.

I understand that having both a `/doc` and a `/docs` directory might seem a bit confusing. The `/docs` directory was chosen because this is the standard folder name that GitHub Pages uses to publish a website from the `master` branch.

We can discuss changing this structure in the future if the current arrangement is inconvenient for the team.

### Next Steps

After this PR is merged, the `s2geometry.io` custom domain can be configured in the repository settings to serve from this `/docs` folder, and the DNS records can be updated.

Thanks!

### DIFF
--- a/docs/CNAME
+++ b/docs/CNAME
@@ -1,0 +1,1 @@
+s2geometry.io


docs/devguide/cpp/examples/term_index.cc
```diff
@@ -0,0 +1,112 @@
+// Copyright 2017 Google Inc. All Rights Reserved.
+// Author: ericv@google.com (Eric Veach)
+//
+// This example shows how to add spatial data to an information retrieval
```
jmr, line 4: We already have https://github.com/google/s2geometry/blob/master/doc/examples/term_index.cc.

These trees need to be unified more carefully.

docs/Gemfile
```diff
@@ -0,0 +1,28 @@
+source "https://rubygems.org"
```
jmr, line 1: why is this `Gemfile` here?

docs/LICENSE
```diff
@@ -0,0 +1,204 @@
+                                 Apache License
```
jmr, line 1: Why is another license being added?

ivan-gerasimov:
> What's with the CNAME changes? Why is CNAME being deleted?

Hi @jmr,

Good question. The CNAME file is managed by GitHub's Pages service. It was automatically deleted when the custom domain in the repository settings was explicitly cleared.

GitHub will auto-generate the CNAME file again once "s2geometry.io" is set as the custom domain in the settings for the repository that will host the live site.

This action is currently blocked because the domain is still associated with the old "manshreck/s2geometry.github.io" repository. After that association is removed, we can configure the domain here and the file will be automatically populated by GitHub.

PR #494, opened 2025-12-14
S2ClosestEdgeQueryBase::InitCovering: Use a single iterator

benguild:
This is sibling to: https://github.com/golang/geo/pull/248
While resolving a bug with the code sync between Go/C++ (https://github.com/golang/geo/pull/247) I noticed the TODO.

Since there's seemingly an effort to mirror implementations, I'm opening a draft PR, because I barely work with C++ and I'd need someone to take a look at this.

### DIFF
--- a/src/s2/s2closest_edge_query_base.h
+++ b/src/s2/s2closest_edge_query_base.h
@@ -305,8 +305,6 @@ class S2ClosestEdgeQueryBase {
   void FindClosestEdgesOptimized(std::optional<ResultVisitor> visitor = {});
   void InitQueue();
   void InitCovering();
-  void AddInitialRange(const S2ShapeIndex::Iterator& first,
-                       const S2ShapeIndex::Iterator& last);
   void MaybeAddResult(const S2Shape& shape, int shape_id, int edge_id);
   void AddResult(const Result& result);
   void ProcessEdges(const QueueEntry& entry);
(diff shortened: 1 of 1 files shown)


jmr:
@ericveach

benguild:
Due to the complexity of this and the Go version, it'd help to have the feedback about whether or not anyone has thoughts on how to improve it, or if the TODO from the current code should just be removed from both Go and C++ otherwise.

I'm happy to take another look but I can't keep this in mind forever and would need to move onto other things if there's no bandwidth to consider the changes.

jmr:
> but I can't keep this in mind forever

If it's going to be a problem for you to come back to the code in a few days or weeks, then it's definitely going to be a problem for someone else to look at it in a few years.  Can you make sure the code is sufficiently documented so you don't need to "keep it in mind"?

I will release the benchmarks soon, so you can test whether this makes a difference.

Funny that it's so much more complex in go than C++. What's that?

src/s2/s2closest_edge_query_base.h
```diff
@@ -921,53 +919,55 @@ void S2ClosestEdgeQueryBase<Distance>::InitCovering() {
   // Don't need to reserve index_cells_ since it is an InlinedVector.
   index_covering_.reserve(6);
 
-  // TODO(ericv): Use a single iterator (iter_) below and save position
```
jmr, line 924: `S2ClosestEdgeQueryBase` has an `iter_` member, which is probably what this TODO is referring to:

https://github.com/google/s2geometry/blob/fd727ebd5c68dfd79d437d579fa988e3ab5cf37d/src/s2/s2closest_edge_query_base.h#L452

jmr:
There are many benchmarks in `s2closest_edge_query_test.cc`.

https://github.com/google/s2geometry/blob/f79c01756c73580ae977308085244e085de8a255/src/s2/s2closest_edge_query_test.cc#L1001

Do we have coverage for this function and does PR speed them up?

benguild:
Closing this draft alongside Go sibling PR, with suggestion the TODO be reconsidered. https://github.com/golang/geo/pull/248#issuecomment-4169003648

PR #496, opened 2025-12-14
Modernize Bazel build configuration with Bazel 9 compatibility and BCR best practices

mmorel-35:
Applied best practices from Bazel Central Registry patches and buildifier linting to standardize build configuration across all BUILD files. Added Bazel 9 compatibility with proper load statements and separated development dependencies. Upgraded dependencies to latest stable versions with upstream fixes.

## Core Fixes

- **Fixed circular dependency**: `s2/util/endian/BUILD` incorrectly referenced `@s2geometry//s2/util/gtl` instead of `//s2/util/gtl`
- **Resolved buildifier warnings**: Moved constant patterns out of glob expressions
- **Standardized formatting**: Applied buildifier to all BUILD files (alphabetical sorting, consistent spacing, trailing commas)

## MODULE.bazel Enhancements

- Added `bazel_skylib` (v1.8.1) and `openssl` (v3.3.1.bcr.1) dependencies
- Updated `rules_cc` to v0.2.14 for Bazel 9 compatibility
- **Upgraded dependencies**: `re2@2025-11-05.bcr.1` and `googletest@1.17.0.bcr.2` (have cc_configure fix built-in)
- **Properly ordered**: Extensions first, then production dependencies alphabetically, then development dependencies alphabetically
- **Separated dev dependencies**: Marked `googletest` and `cc_configure_extension` as `dev_dependency = True`
- **Removed workaround**: Issue bazelbuild/bazel#24426 is now fixed upstream in re2, googletest, and abseil-cpp

## Bazel 9 Compatibility

- Added `load("@rules_cc//cc:defs.bzl", ...)` statements to all BUILD files (required in Bazel 9)
- Updated `bazel_compatibility` to support Bazel 7.0.0 through 9.x
- Verified builds work with both Bazel 8.5.0 and 9.0.0rc3
- Removed `includes = [""]` from main s2 library (causes errors in Bazel 9)

## Feature: SSL Library Selection

Added build-time selection between BoringSSL and OpenSSL:

```bash
# Default (BoringSSL)
bazel build //:s2

# With OpenSSL
bazel build //:s2 --//:use_openssl
```

Implementation via bool_flag in `BUILD.bazel` and select() in `s2/util/math/exactfloat/BUILD`:

```starlark
deps = [
    "//s2/base:logging",
    "@abseil-cpp//absl/log:absl_check",
    "@abseil-cpp//absl/log:log",
] + select({
    "//:use_openssl_setting": ["@openssl//:crypto"],
    "//conditions:default": ["@boringssl//:crypto"],
})
```

## Additional Improvements

- Removed redundant default parameters from glob calls
- Consistent attribute ordering (srcs before hdrs) across all targets

### DIFF
--- a/src/MODULE.bazel
+++ b/src/MODULE.bazel
@@ -7,15 +7,16 @@ module(
     compatibility_level = 1,
 )
 
-cc_configure = use_extension("@rules_cc//cc:extensions.bzl", "cc_configure_extension")
-use_repo(cc_configure, "local_config_cc")
-
-bazel_dep(name = "boringssl", version = "0.20250514.0")
+# Production dependencies
 bazel_dep(name = "abseil-cpp", version = "20250814.1")
-bazel_dep(name = "googletest", version = "1.17.0")
-bazel_dep(name = "rules_cc", version = "0.1.1")
+bazel_dep(name = "bazel_skylib", version = "1.8.1")
+bazel_dep(name = "boringssl", version = "0.20250514.0")
+bazel_dep(name = "openssl", version = "3.3.1.bcr.1")
+bazel_dep(name = "rules_cc", version = "0.2.14")
 
-# Forces usage of the patched re2 version that marks cc_configure_extension as a
-# dev dep as a workaround for https://github.com/bazelbuild/bazel/issues/24426.
-# TODO: remove once fixed upstream in re2, googletest, and abseil-cpp.
-bazel_dep(name = "re2", version = "2024-07-02.bcr.1")
+# Development dependencies (only needed for tests)
+bazel_dep(name = "googletest", version = "1.17.0.bcr.2", dev_dependency = True)
+
+# Extensions
+cc_configure = use_extension("@rules_cc//cc:extensions.bzl", "cc_configure_extension", dev_dependency = True)
+use_repo(cc_configure, "local_config_cc")
--- a/src/s2/BUILD
+++ b/src/s2/BUILD
@@ -1,19 +1,29 @@
 package(default_visibility = ["//visibility:public"])
-exports_files(glob(
-  include = [
-    "internal/*.h", "internal/*.cc",
-    "r1*.h","r1*.cc",
-    "r2*.h","r2*.cc",
-    "s1*.h","s1*.cc",
-    "s2*.h","s2*.cc",
-    "encoded*.h","encoded*.cc",
-    "*lexicon.h","*lexicon.cc",
-    "_fp*.h","_fp*.cc",
-    "mutable*.h","mutable*.cc",
-    "thread_testing.h","thread_testing.cc",
-    "*test.h","*test.cc",
-  ], 
-  exclude=[], 
-  exclude_directories=0, 
-  allow_empty=True)
+
+exports_files(
+    glob(
+        [
+            "*lexicon.h",
+            "*lexicon.cc",
+            "*test.cc",
+            "_fp*.h",
+            "encoded*.h",
+            "encoded*.cc",
+            "internal/*.h",
+            "internal/*.cc",
+            "mutable*.h",
+            "mutable*.cc",
+            "r1*.h",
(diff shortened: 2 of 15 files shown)


diodoliz:
Thanks for this change, I was about to start on this as well.

Without `load("@rules_cc//cc:defs.bzl", "cc_library")` builds with bazel 10 (rolling) are completely broken.

https://github.com/bazelbuild/bazel-central-registry/pull/6822#issuecomment-3650679162

src/s2/BUILD
```diff
@@ -1,19 +1,32 @@
...
+            "*lexicon.h",
+            "*lexicon.cc",
+            "_fp*.h",
+            "_fp*.cc",
+            "mutable*.h",
+            "mutable*.cc",
+            "*test.h",
+            "*test.cc",
+        ],
+        allow_empty = True,
```
jmr: do we need to `allow_empty`?

mmorel-35: From https://bazel.build/reference/be/functions : 
If the `allow_empty` argument is set to `False`, the `glob` function will error-out if the result would otherwise be the empty list.

src/MODULE.bazel
```diff
@@ -7,15 +7,17 @@ module(
...
+bazel_dep(name = "openssl", version = "3.3.1.bcr.1")
+bazel_dep(name = "re2", version = "2025-11-05.bcr.1")
+bazel_dep(name = "rules_cc", version = "0.2.14")
+
+# Development dependencies (only needed for tests)
+bazel_dep(name = "googletest", version = "1.17.0.bcr.2", dev_dependency = True)
 
-# Forces usage of the patched re2 version that marks cc_configure_extension as a
-# dev dep as a workaround for https://github.com/bazelbuild/bazel/issues/24426.
-# TODO: remove once fixed upstream in re2, googletest, and abseil-cpp.
```
jmr, line 20: The TODO said to remove this. Is that out of date now?

@vtsao-openai

src/s2/BUILD
jmr: Yes, and is this going to be an error? Which globs don't match?

src/MODULE.bazel
jmr: It was a question, and I don't know the answer.  Which of these do we need/want and why?

src/s2/BUILD
jmr: It saves time if you test before you push.

https://github.com/google/s2geometry/actions/runs/20227386992/job/58067018647?pr=496#step:4:21

If these patterns don't match, what are they for?

src/s2/BUILD
```diff
@@ -1,19 +1,29 @@
...
+            "r2*.cc",
+            "s1*.h",
+            "s1*.cc",
+            "s2*.h",
+            "s2*.cc",
+            "encoded*.h",
+            "encoded*.cc",
+            "*lexicon.h",
+            "*lexicon.cc",
+            "_fp*.h",
```
jmr: Please sort

src/MODULE.bazel
jmr: Ok, so it looks like it's right to remove this and we should only declare direct dependencies or indirect ones where we need to override a version.

src/s2/BUILD
```diff
@@ -1,19 +1,29 @@
...
-    "_fp*.h","_fp*.cc",
-    "mutable*.h","mutable*.cc",
-    "thread_testing.h","thread_testing.cc",
-    "*test.h","*test.cc",
-  ], 
-  exclude=[], 
-  exclude_directories=0, 
-  allow_empty=True)
+
+exports_files(
```
jmr, line 3: Wait, so while we're modernizing, why do we have a BUILD file that just exports everything?  Would it make more sense to just delete `s2/BUILD`?

We can leave this for future cleanups.

Issue #522, opened 2026-01-06
Pybind wrapper for s2


deustis:
The repo README indicates future support for pybind is planned but I wasn't able to find an issue tracking this --

https://github.com/google/s2geometry/tree/master?tab=readme-ov-file#apiabi-stability
> The Python API is particularly unstable, and it is planned that the SWIGged API will be replaced by a pybind11 version with more Pythonic names and more complete functionality.

Any indication on when this might become available?

rwgk:
> Thanks. Is there a good way to avoid maintaining both clif bindings and pybind11/nanobind ones? Is the clif output close to usable as a starting point for pybind11 bindings?

I don't think so, but I don't know what the current state is of clif inside Google (I switched companies).

These days, I'd probably just ask my favorite LLM to convert the clif bindings to pybind11 or nanobind. I expect it'll probably do a reasonable job that can be finished with a small amount of manual fix-ups.

Maybe you can help it by giving the LLM the output of `help(yourmodule)`, so that it can "see" the built bindings.

deustis:
We happen to use pybind11 at my current company (Aurora Innovation) and it works well for us! I'd be happy with any of those options though. I was mainly curious to understand if I should be avoiding the current Python SWIG wrappers because some new/better Python API is forthcoming.

deustis:
BTW, I'd be open to contributing a PR for the pybind option if that helps. I'm less familiar with the other Python binding libraries but could give any of those a shot if they are preferred.

BTW, former Googler here. I worked on Google Maps from 2005 to 2013! Jesse you're name looks really familiar although I don't recall if we crossed paths!

deustis:
@jmr, are you in touch with the maintainer of the current PyPI `s2geometry` package?:
https://pypi.org/project/s2geometry/

The maintainer is listed as @figroc:
https://github.com/figroc

We're getting close to feature parity on the SWIG bindings, and I'm thinking we'll want to publish the new library as a new major version of that `s2geometry` package.

figroc:
@deustis I've added this repo as a Trusted Publisher in #406.

jmr:
> We're getting close to feature parity on the SWIG bindings, and I'm thinking we'll want to publish the new library as a new major version of that `s2geometry` package.

It doesn't look like we're that close, but getting there. We should have a compat layer to avoid breaking everyone.

https://github.com/google/s2geometry/issues/583

I would also also give the new code some bake time before making a release.

PR #524, opened 2026-01-08
pybind: Add S2Point bindings

deustis:
Initial Bazel configuration for Pybind11 and S2Point bindings.
https://github.com/google/s2geometry/issues/522

The S2 Geometry library is transitioning from SWIG-based bindings to pybind11-based bindings.

During this migration:
- **SWIG bindings** (`s2geometry`): The current production bindings, built with CMake. Use `import s2geometry` to access these.
- **pybind11 bindings** (`s2geometry_pybind`): The new bindings under development, built with Bazel. Use `import s2geometry_pybind` to access these.

Once the pybind11 bindings are feature-complete and stable, the SWIG bindings will be deprecated and the pybind11 package will be renamed to `s2geometry` to become the primary Python API.

This PR adds Bazel configuration for building and testing the bindings on the local toolchain. For distribution we'll need to generate a wheel for compatibility with different OS and python versions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 bazel-*
+build/
 MODULE.bazel.lock
--- a/src/MODULE.bazel
+++ b/src/MODULE.bazel
@@ -18,6 +18,13 @@ bazel_dep(name = "openssl", version = "3.5.4.bcr.0", dev_dependency = True)
 bazel_dep(name = "google_benchmark", version = "1.9.4", dev_dependency = True)
 bazel_dep(name = "googletest", version = "1.17.0.bcr.2", dev_dependency = True)
 
+# Python configuration
+# The platforms dependency is needed for pybind_extension rules. See:
+# https://github.com/pybind/pybind11_bazel/blob/master/MODULE.bazel
+bazel_dep(name = "platforms", version = "1.0.0")
+bazel_dep(name = "pybind11_bazel", version = "3.0.0")
+bazel_dep(name = "rules_python", version = "1.6.3")
+
 # Extensions
 cc_configure = use_extension("@rules_cc//cc:extensions.bzl", "cc_configure_extension", dev_dependency = True)
 use_repo(cc_configure, "local_config_cc")
--- a/src/python/README.md
+++ b/src/python/README.md
@@ -1,0 +1,85 @@
+# S2 Geometry Pybind11 Bindings
+
+*UNDER DEVELOPMENT*
+
+This directory contains SWIG and pybind11-based Python SDK for S2 Geometry.
+
+## Migration Strategy
+
+The S2 Geometry library is transitioning from SWIG-based bindings to pybind11-based bindings. During this migration:
+
+- **SWIG bindings** (`s2geometry`): The current production bindings, built with CMake. Use `import s2geometry` to access these.
+- **pybind11 bindings** (`s2geometry_pybind`): The new bindings under development, built with Bazel. Use `import s2geometry_pybind` to access these.
+
+Once the pybind11 bindings are feature-complete and stable, the SWIG bindings will be deprecated and the pybind11 package will be renamed to `s2geometry` to become the primary Python API.
+
+## Directory Structure
+
+```
+python/
+├── module.cc                     # Binding module entry point
+├── s2point_bindings.cc           # Bindings for S2Point (add more *_bindings.cc as needed)
+├── s2geometry_pybind/            # Dir for Python package
+│   └── __init__.py               # Package initialization
+├── s2point_test.py               # Tests for S2Point (add more *_test.py as needed)
+└── BUILD.bazel                   # Build rules for bindings, library, and tests
+```
+
+## Usage Example
+
+```python
+import s2geometry_pybind as s2
+
+p1 = s2.S2Point(1.0, 0.0, 0.0)
+p2 = s2.S2Point(0.0, 1.0, 0.0)
+sum_point = p1 + p2
+print(sum_point)
+```
+
+## Development
+
+### Building with Bazel (pybind11 bindings)
+
+Bazel can be used for development and testing of the new pybind11-based bindings.
+
+To run all tests:
+```bash
+cd src
+bazel test //python/...
+```
+
+### Building with CMake (SWIG bindings)
+
+CMake currently builds **only the SWIG-based bindings** (the legacy `s2geometry` package). The pybind11 bindings are not yet integrated into the CMake build system.
+
+For detailed CMake build instructions, dependency installation, and Python wheel creation, see the [parent directory README](../README.md#build-and-install). Key points:
+
+- Install dependencies: `sudo apt-get install cmake libssl-dev swig python3-dev`
+- Enable Python bindings with `-DWITH_PYTHON=ON` when running cmake
+- This will build the SWIG-based `s2geometry` package only
+
+Example:
+```bash
+mkdir build && cd build
+cmake -DBUILD_TESTS=yes -DWITH_PYTHON=ON -DCMAKE_PREFIX_PATH=/path/to/absl/install -DCMAKE_CXX_STANDARD=17 ..
+make -j $(nproc)
+make test ARGS="-j$(nproc)"
+sudo make install
+```
+
+To run the SWIG tests directly without cmake:
+```bash
+cd build/python
+PYTHONPATH=. python3 ../../src/python/s2geometry_test.py
+```
+
+**Note:** Once the pybind11 bindings are complete, they will be integrated into the CMake build system as a replacement for the SWIG bindings.
+
+## Extending the Bindings
+
+To add bindings for a new class:
+
+1. Create `<classname>_bindings.cc` with pybind11 bindings
+2. Update `BUILD.bazel` to add a new `pybind_library` target
+3. Update `module.cc` to call your binding function
+4. Create tests in `<classname>_test.py`
--- a/AUTHORS
+++ b/AUTHORS
@@ -8,6 +8,7 @@
 #
 # Please keep the list sorted.
 
+Aurora Innovation
 Dan Larkin-York <dan@arangodb.com>
 Google Inc.
 Koordinates Limited
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -23,6 +23,7 @@
 # Please keep the list sorted.
 
 Dan Larkin-York <dan@arangodb.com>
+David Eustis <deustis@aurora.tech>
 Eric Veach <ericv@google.com>
 Jesse Rosenstock <jmr@google.com>
 Julien Basch <julienbasch@google.com>
--- a/src/python/BUILD.bazel
+++ b/src/python/BUILD.bazel
@@ -1,0 +1,49 @@
+load("@rules_cc//cc:defs.bzl", "cc_library")
+load("@pybind11_bazel//:build_defs.bzl", "pybind_extension", "pybind_library")
+load("@rules_python//python:defs.bzl", "py_library", "py_test")
+
+package(default_visibility = ["//visibility:private"])
+
+# ========================================
+# Python Package
+# ========================================
+
+py_library(
+    name = "s2geometry_pybind",
+    srcs = ["s2geometry_pybind/__init__.py"],
+    data = [
+        ":s2geometry_bindings",
+    ],
+    imports = ["."],
+    visibility = ["//visibility:public"],
+)
+
+# ========================================
+# C++ Bindings
+# ========================================
+
+pybind_extension(
+    name = "s2geometry_bindings",
+    srcs = ["module.cc"],
+    deps = [
+        ":s2point_bindings",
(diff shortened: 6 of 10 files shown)


src/MODULE.bazel
```diff
@@ -18,6 +18,11 @@ bazel_dep(name = "rules_cc", version = "0.2.16")
 bazel_dep(name = "google_benchmark", version = "1.9.4", dev_dependency = True)
 bazel_dep(name = "googletest", version = "1.17.0.bcr.2", dev_dependency = True)
 
+# Python configuration
+bazel_dep(name = "platforms", version = "1.0.0")
+bazel_dep(name = "rules_python", version = "1.6.3")
```
jmr: sort

src/python/PYBIND_README.md
```diff
@@ -0,0 +1,62 @@
+# S2 Geometry Pybind11 Bindings
+
+*UNDER DEVELOPMENT*
+
+This directory contains pybind11-based Python SDK for S2 Geometry.
```
jmr: SWIG- and pybind11-based

src/python/PYBIND_README.md
```diff
@@ -0,0 +1,62 @@
...
+
+p1 = s2.S2Point(1.0, 0.0, 0.0)
+p2 = s2.S2Point(0.0, 1.0, 0.0)
+sum_point = p1 + p2
+print(sum_point)
+```
+
+## Development
+
+Bazel can be used for development and testing.
```
jmr: What's the CMake / pybind11 story? Say something about that here.

src/python/README.md
```diff
@@ -0,0 +1,62 @@
...
+- **SWIG bindings** (`s2geometry`): The current production bindings, built with CMake. Use `import s2geometry` to access these.
+- **pybind11 bindings** (`s2geometry_pybind`): The new bindings under development, built with Bazel. Use `import s2geometry_pybind` to access these.
+
+Once the pybind11 bindings are feature-complete and stable, the SWIG bindings will be deprecated and the pybind11 package will be renamed to `s2geometry` to become the primary Python API.
+
+## Directory Structure
+
+```
+python/
+├── bindings/                     # Dir for C++ pybind11 bindings
```
jmr: Why is there both `bindings/` and `s2geometry_pybind/`?

It seems like one should do, and everything could probably go directly in `python/`, but I'm not sure.

src/python/README.md
jmr, line 1: Can you just call it `README.md`?

src/python/tests/s2point_test.py
```diff
@@ -0,0 +1,120 @@
...
+
+import unittest
+import s2geometry_pybind as s2
+
+
+class TestS2Point(unittest.TestCase):
+    """Test cases for S2Point bindings."""
+
+    def test_default_constructor(self):
+        """Test default constructor creates zero point."""
```
jmr: These docstrings seem like ` x += 1  // add one to x` comments.

src/python/tests/s2point_test.py
```diff
@@ -0,0 +1,120 @@
...
+        p3 = s2.S2Point(1.0, 2.0, 4.0)
+        self.assertTrue(p1 == p2)
+        self.assertFalse(p1 == p3)
+        self.assertTrue(p1 != p3)
+
+    def test_repr(self):
+        """Test string representation."""
+        p = s2.S2Point(1.0, 2.0, 3.0)
+        repr_str = repr(p)
+        self.assertIn("S2Point", repr_str)
```
jmr: I think it would be good to put the exact string so we can see what that is.

src/python/tests/s2point_test.py
```diff
@@ -0,0 +1,120 @@
...
+        self.assertAlmostEqual(result.x(), -1.0)
+        self.assertAlmostEqual(result.y(), 2.0)
+        self.assertAlmostEqual(result.z(), -3.0)
+
+    def test_equality(self):
+        """Test point equality."""
+        p1 = s2.S2Point(1.0, 2.0, 3.0)
+        p2 = s2.S2Point(1.0, 2.0, 3.0)
+        p3 = s2.S2Point(1.0, 2.0, 4.0)
+        self.assertTrue(p1 == p2)
```
jmr: why not `assertEqual` / `assertNotEqual`?

src/python/README.md
```diff
@@ -0,0 +1,62 @@
...
+│   ├── BUILD.bazel               # Binding build rules
+│   ├── module.cc                 # Binding module entry point
+│   ├── s2point_bindings.cc       # Bindings for S2Point
+│   └── ...                       # Bindings for additional classes
+├── s2geometry_pybind/            # Dir for Python package
+│   └── __init__.py               # Package initialization
+├── tests/                        # Dir for Python unit tests
```
jmr: `tests/` subdir also seems inconsistent with the rest of the project. Why not keep tests with the code?

src/python/bindings/BUILD.bazel
```diff
@@ -0,0 +1,20 @@
+load("@rules_cc//cc:defs.bzl", "cc_library")
+load("@pybind11_bazel//:build_defs.bzl", "pybind_extension", "pybind_library")
+
+package(default_visibility = ["//python:__subpackages__"])
+
+pybind_extension(
+    name = "s2geometry_bindings",
```
jmr: Should this be public?

src/python/bindings/BUILD.bazel
```diff
@@ -0,0 +1,20 @@
...
+    srcs = ["module.cc"],
```
jmr: Does it make any sense to have fine-grained targets?  In google3, we have fine-grained targets for the S2 library, which really helps compilation and development time.  We don't use them for PyClif mostly because the C++ targets were sharded after the pyclif ones were added.

Do we just accept that it's monolithic in Python?

src/python/tests/s2point_test.py
```diff
@@ -0,0 +1,120 @@
...
+import s2geometry_pybind as s2
+
+
+class TestS2Point(unittest.TestCase):
+    """Test cases for S2Point bindings."""
+
+    def test_default_constructor(self):
+        """Test default constructor creates zero point."""
+        p = s2.S2Point()
+        self.assertEqual(p.x(), 0.0)
```
jmr: In the pyclif bindings, we made `.x` a property, under the assumption that this is how it would have been in pure Python. Is that possible in pybind11?  Which do you think is more Pythonic?

jmr (changes requested):

src/python/README.md
deustis: I believe we do do need the `s2geometry_pybind` to match the name of the python package. Open to suggestions on naming there.

The `bindings` and `tests` directories are not necessary. We can certainly flatten one or both of those into the `python` parent if you prefer. Happy to go with whatever layout you prefer here.

src/python/README.md
deustis: Sure, we can drop. Just let me know what layout you prefer in the comment above.

src/MODULE.bazel
deustis: Done

src/python/tests/s2point_test.py
deustis: Removed!

src/python/tests/s2point_test.py
deustis: Done

src/python/README.md
deustis: Done

src/python/PYBIND_README.md
deustis: Done

src/python/tests/s2point_test.py
deustis: Good call, updated.

src/python/PYBIND_README.md
deustis: Added some explanation about this. Currently CMake only builds the SWIG package. We'll probably need to update CMake to build the wheel for the final distribution but I'll save that for another PR if that's alright with you?

src/python/bindings/BUILD.bazel
deustis: We could consider adding some submodules. How about if we put the "core" classes directly under the top-level package (e.g. `s2geometry.S2Cell`), but add submodules for some of the utility classes (e.g. `s2geometry.regions.S2CellCovering`). Open to suggestions on how to break this up.

Once we align on the python module structure, I think we would then define one `pybind_extension` rule per module, plus separate `pybind_library` rules for each class.

src/python/bindings/BUILD.bazel
deustis: Add the underscore prefix to indicate it is a private module

src/python/PYBIND_README.md
jmr: Yes, another PR SGTM.

src/MODULE.bazel
```diff
@@ -18,6 +18,11 @@ bazel_dep(name = "rules_cc", version = "0.2.16")
 bazel_dep(name = "google_benchmark", version = "1.9.4", dev_dependency = True)
 bazel_dep(name = "googletest", version = "1.17.0.bcr.2", dev_dependency = True)
 
+# Python configuration
+bazel_dep(name = "platforms", version = "1.0.0")
```
jmr, line 24: I don't see where this is used.

src/python/README.md
jmr: Well, definitely flatten the extra dirs.

If it needs to be `s2geometry_pybind`, then so be it.  If not, we can further flatten later.

What will be the bazel targets and what will people import in Python?

src/python/tests/s2point_test.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def test_negation(self):
+        p = s2.S2Point(1.0, -2.0, 3.0)
+        result = -p
+        self.assertAlmostEqual(result.x, -1.0)
+        self.assertAlmostEqual(result.y, 2.0)
+        self.assertAlmostEqual(result.z, -3.0)
+
+    def test_equality(self):
+        """Test point equality."""
```
jmr: there are still some low-content comments

src/MODULE.bazel
deustis: I believe this is needed for pybind.

Without this I get errors like:
```
ERROR: no such package '@@[unknown repo 'platforms' requested from @@]//os': The repository '@@[unknown repo 'platforms' requested from @@]' could not be resolved: No repository visible as '@platforms' from main repository
ERROR: /home/deustis/src/s2geometry/src/python/bindings/BUILD.bazel:6:17: errors encountered resolving select() keys for //python/bindings:_s2geometry_bindings.so
```

I see it declared upstream here, but I'm not entirely sure it also needs to be added as a dependency in this module:
https://github.com/pybind/pybind11_bazel/blob/master/MODULE.bazel#L9

I've added a comment about this.

src/python/tests/s2point_test.py
deustis: Fixed

src/python/tests/s2point_test.py
deustis: Done

src/python/README.md
deustis: I've flattened the directories!

> What will be the bazel targets and what will people import in Python?

I'm thinking the bazel targets will be mainly for development but I suppose for a production use-case if a customer is already using bazel it should be possible to depend on the `py_library` target directly (`s2geometry_pybind`).

That said, I think most users would want a pip package so I mentioned we'll need to figure out how to build a wheel and upload to pypi.

Currently the python import is `s2geometry_pybind` as seen in the unit test:
https://github.com/google/s2geometry/pull/524/changes#diff-f1182576e6dc7943a0970fce1cdeee021cea55a7b88b9c50756d6cb9fa8d4126R4

I was thinking that once we have a full replacement for the SWIG version we will rename to `s2geometry` (and delete the former)

.gitignore
```diff
@@ -1,2 +1,4 @@
+.cache/
```
jmr: what makes .cache/?

src/python/bindings/BUILD.bazel
jmr: Can you use `visibility`?

src/python/bindings/BUILD.bazel
jmr: I'm not sure how much sense "regions" makes.  We can forget this for now and think again when the external BUILD.bazel targets are sharded.

jmr (commented):
Looks good modulo the two new comments/questions.

Oh, also add the CONTRIBUTORS stuff here.

.gitignore
deustis: Dropped. Must have been something unrelated.

src/python/bindings/BUILD.bazel
deustis: Done

deustis:
> Oh, also add the CONTRIBUTORS stuff here.

Done!

PR #534, opened 2026-01-28
pybind: Add S1Interval bindings (#524)

deustis:
Add pybind11 bindings for S1Interval.

Also add interface notes to the python README

(Part of a series of addressing #524)

### DIFF
--- a/src/python/README.md
+++ b/src/python/README.md
@@ -44,6 +44,9 @@ The Python bindings follow the C++ API closely but with Pythonic conventions:
 - Invalid inputs to constructions or functions raises `ValueError`.
 - Example: `S1Interval(0.0, 4.0)` raises `ValueError` because `4.0 > π`.
 - Note: In C++, these conditions trigger `ABSL_DCHECK` assertions. The bindings prevent these assertions from firing by pre-validating inputs.
+- Note: Python bindings check for invalid inputs and throw C++ exceptions which are caught by
+  pybind and converted to Python exceptions. Exceptions are normally prohibited by the C++
+  style guide, but this is the preferred approach for pybind.
 
 **Documentation:**
 - Python docstrings provide essential information about parameters, return values, and key behaviors
--- a/src/python/s1interval_bindings.cc
+++ b/src/python/s1interval_bindings.cc
@@ -10,7 +10,9 @@ namespace py = pybind11;
 namespace {
 
 void MaybeThrowInvalidPoint(double p) {
-  if (!S1Interval::IsValidPoint(p)) throw py::value_error("Invalid S1 point: " + std::to_string(p));
+  if (!S1Interval::IsValidPoint(p)) {
+     throw py::value_error("Invalid S1 point: " + std::to_string(p));
+  }
 }
 
 }  // namespace
@@ -129,8 +131,10 @@ void bind_s1interval(py::module& m) {
            "by at most 'max_error' (radians) to match the other interval.")
 
       // Operators
-      .def(py::self == py::self, "Return true if two intervals contain the same set of points")
-      .def(py::self != py::self, "Return true if two intervals do not contain the same set of points")
+      .def(py::self == py::self,
+          "Return true if two intervals contain the same set of points")
+      .def(py::self != py::self,
+          "Return true if two intervals do not contain the same set of points")
 
       // String representation
       .def("__repr__", [](const S1Interval& i) {


deustis:
@jmr, next installment here. I've added some updates to the README that explain some of the user-facing interface choices for the python bindings. That might be a good place to start for the review. Feel free to suggest any changes.

src/python/BUILD.bazel
```diff
@@ -25,14 +25,26 @@ py_library(
 pybind_extension(
     name = "s2geometry_bindings",
     srcs = ["module.cc"],
+    copts = ["-DNDEBUG"],
```
jmr: I'm not sure about this. Definitely needs a comment.

src/python/README.md
```diff
@@ -36,8 +26,63 @@ sum_point = p1 + p2
 print(sum_point)
 ```
 
+### Interface Notes
+
+The Python bindings follow the C++ API closely but with Pythonic conventions:
+
+**Naming Conventions:**
+- Core classes exist within the top-level module; we intend to define submodules for utility classes.
```
jmr: What would these submodules look like?  Maybe leave it as "may define", since I don't really know.

src/python/README.md
```diff
@@ -36,8 +26,63 @@ sum_point = p1 + p2
...
+- Class names remain unchanged (e.g., `S2Point`, `S1Angle`, `R1Interval`)
+- Method names are converted to snake_case (converted from UpperCamelCase C++ function names)
+
+**Properties vs. Methods:**
+- Simple coordinate accessors are properties: `point.x`, `point.y`, `interval.lo`, `interval.hi`
+- If the C++ class has a trivial set_foo method for the property, then the Python property is mutable (otherwise it is a read-only property).
+- Other functions are not properties: `angle.radians()`, `angle.degrees()`, `interval.get_length()`
```
jmr: Add `get_` prefix is removed.

src/python/README.md
```diff
@@ -36,8 +26,63 @@ sum_point = p1 + p2
...
+
+**Invalid/Unnormalized Values:**
+- Some constructors accept invalid values or unnormalized values.
+- Examples: 
+  - `S1Interval(0.0, 4.0)` with `4.0 > π` creates interval where `is_valid()` is `False`
+  - `S2Point(3.0, 4.0, 0.0)` creates an unnormalized point outside of the unit sphere; use `normalize()` to normalize.
+- In **C++** invalid values will typically trigger (`ABSL_DCHECK`) when compiled with debug options.
+- In **Python bindings** debug assertions (`ABSL_DCHECK`) are disabled, matching the production optimized C++ behavior.
```
jmr: Hmm. Why? Isn't this an ODR violation waiting to happen?

src/python/README.md
```diff
@@ -36,8 +26,63 @@ sum_point = p1 + p2
...
+
+**Type Conversions:**
+- Python floats map to C++ `double`
```
jmr: This seems obvious.

src/python/README.md
```diff
@@ -82,4 +127,18 @@ To add bindings for a new class:
...
 3. Update `module.cc` to call your binding function
-4. Create tests in `<classname>_test.py`
\ No newline at end of file
+4. Create tests in `<classname>_test.py`
+
+### Binding File Organization
+
+Use the following sections to organize functions within the bindings files and tests:
+
+1. **Constructors** - Default constructors and constructors with parameters
```
jmr, line 138: This seems reasonable, but it's also helpful if the order reflects the C++ order.

src/python/s1interval_bindings.cc
```diff
@@ -0,0 +1,96 @@
...
+
+      // Predicates
+      .def("is_valid", &S1Interval::is_valid, "Check if interval is valid")
+      .def("is_full", &S1Interval::is_full, "Check if interval is full")
+      .def("is_empty", &S1Interval::is_empty, "Check if interval is empty")
+      .def("is_inverted", &S1Interval::is_inverted,
+           "Check if interval is inverted (lo > hi)")
+
+      // Geometric operations
+      .def("get_center", &S1Interval::GetCenter, "Return center of interval")
```
jmr: no `get_`

src/python/s1interval_bindings.cc
```diff
@@ -0,0 +1,96 @@
...
+           "Return intersection with another interval")
+      .def("complement", &S1Interval::Complement,
+           "Return complement of this interval")
+      .def("get_directed_hausdorff_distance", 
+           &S1Interval::GetDirectedHausdorffDistance,
+           py::arg("other"),
+           "Return directed Hausdorff distance to another interval")
+      .def("approx_equals", &S1Interval::ApproxEquals,
+           py::arg("other"), py::arg("max_error") = 1e-15,
+           "Check if approximately equal")
```
jmr: Let's make the docstrings more informative. `max_error` is radians, etc.

Can you tell your LLM to copy the docstrings from C++?

src/python/s1interval_bindings.cc
```diff
@@ -0,0 +1,96 @@
...
+      .def("intersection", &S1Interval::Intersection, py::arg("other"),
+           "Return intersection with another interval")
+      .def("complement", &S1Interval::Complement,
+           "Return complement of this interval")
+      .def("get_directed_hausdorff_distance", 
+           &S1Interval::GetDirectedHausdorffDistance,
+           py::arg("other"),
+           "Return directed Hausdorff distance to another interval")
+      .def("approx_equals", &S1Interval::ApproxEquals,
+           py::arg("other"), py::arg("max_error") = 1e-15,
```
jmr, line 128: Is there a way for pybind to get the default from C++?

src/python/s1interval_bindings.cc
```diff
@@ -0,0 +1,96 @@
...
+      .def(py::self == py::self, "Check equality")
+      .def(py::self != py::self, "Check inequality")
+
+      // String representation
+      .def("__repr__", [](const S1Interval& i) {
+        return "S1Interval(" + std::to_string(i.lo()) + ", " +
+               std::to_string(i.hi()) + ")";
+      })
+      .def("__str__", [](const S1Interval& i) {
+        if (i.is_empty()) return std::string("[∅]");
```
jmr: Maybe "[empty]"/"[full]" are more consistent with other places.

I also wonder if we should just delegate to operator<< in C++.

jmr (changes requested):

src/python/README.md
deustis: The challenge is that the C++ implementation uses debug assertions which is perhaps the source of the "ODR violation". That leaves us two choices for the Python interface: adhere to the dbg behavior or the opt behavior.

The current PR does the latter, meaning anywhere the opt C++ code accepts invalid inputs the Python does the same (as well as exposing an is_valid if that is available in C++).

Alternatively, we could follow the dbg behavior where the user would expect the Python code to raise exceptions. The challenge is that the ABSL debug checks abort the process and aren't "catchable" in the Python bindings. So we would need to either change the C++ assertions, or reproduce the validation in the Pybind implementation itself.

From the C++ perspective, would you say that the debug assertions are enough of a deterrent that we can say the opt behavior is the de-facto interface?

In that case I agree we should probably lean into the cbg behavior. We might also consider switching from `ABSL_DCHECK` to `ABSL_CHECK` in the C++ code? I'm guessing we can't make larger interface changes to the C++ code such as raising exceptions or returning Status messages.

So that leaves us with trying to reproduce the validity checks in the Python bindings themselves. This is doable but it would be non-trivial For example, for S1Interval we would to refactor this validity function to be able to call it before calling the various C++ function that would fail with assertions:
https://github.com/google/s2geometry/blob/master/src/s2/s1interval.h#L247
```
inline bool S1Interval::is_valid() const {
  return (std::fabs(lo()) <= M_PI && std::fabs(hi()) <= M_PI &&
          !(lo() == -M_PI && hi() != M_PI) &&
          !(hi() == -M_PI && lo() != M_PI));
}
```

src/python/BUILD.bazel
deustis: Agree, let's align on this thread and then I'll update this:
https://github.com/google/s2geometry/pull/534#discussion_r2737488836

src/python/README.md
deustis: Changed to "may define". Agreed let's discuss when we get there.

src/python/README.md
deustis: Removed this and the next line about radians.

src/python/README.md
deustis: Add a note to secondarily follow the ordering in the C++ headers.

src/python/s1interval_bindings.cc
deustis: Done here and below

src/python/s1interval_bindings.cc
deustis: Unfortunately I don't think that is possible. Added some comments here and in the C++ header to keep these in sync.

src/python/s1interval_bindings.cc
deustis: Added more detail to the docstrings here and in S2Point. For longer docstrings I'm using the multi-line convention with a summary line and additional details:
https://peps.python.org/pep-0257/#multi-line-docstrings

I didn't literally copy the C++ comments but I asked the LLM to pull in all of the important details. Also added a note to the README about C++ comments being the source of truth.

src/python/s1interval_bindings.cc
deustis: Updated to use the << operator. __repr__ still prefixes the class name but otherwise the output is the same.

src/python/README.md
jmr: > That leaves us two choices for the Python interface: adhere to the dbg behavior or the opt behavior.

Why can't the user choose by setting the compilation mode? 

> From the C++ perspective, would you say that the debug assertions are enough of a deterrent that we can say the opt behavior is the de-facto interface?

It's the same interface for dbg and opt. The interface is that there are preconditions and if they are violated, anything could happen. This is fine for C++ but is not very Pythonic.

Now I see that `S1Interval` is a bit odd in that it `DCHECK`s in the constructor rather just than when you try to use it. I think we could add `S1Interval::FromBounds(lo, hi) -> optional<S1Interval>` or `StatusOr<S1Interval>` (not sure how much the `Status` would add, but maybe good for consistency). We could also move the `DCHECK` out of the constructor.

The Python constructor could call this and throw `ValueError`, then we deal with only valid intervals.  Maybe the setters need to go to preserve "always valid".

deustis: > The interface is that there are preconditions and if they are violated, anything could happen.

😆 -- IMHO, not so great for C++ either.

> This is fine for C++ but is not very Pythonic.

Agreed! Let's go with exceptions for the Python interface then? I'd also rather avoid differences in behavior depending on compilation mode.

> Now I see that S1Interval is a bit odd in that it DCHECKs in the constructor 

Thanks for the suggestions. I'll need to circle back to this next week but something along those lines should work.

deustis:
Hi there, just a quick note -- I was traveling last week and working against some deadlines this week but still intending to circle back on this!

> Is there a reason you picked S1Interval? Seems obscure and low level.

No particular reason... starting with the lowest level primitives and working my way up to `S2Cell`.

deustis:
@jmr, apologies for the delay. I'm back and have updated the the PR as discussed.

The python interface will now raise value errors for invalid inputs to avoid any possibility of triggering the DCHECKS. In the case of S1Angle this basically means ensuring all of the points are within +/- pi. I've added a static helper to the C++ code to make this easier so we aren't having to do any math in the bindings themselves.

src/s2/s1interval.cc
```diff
@@ -69,14 +69,14 @@ bool S1Interval::FastContains(double p) const {
 
 bool S1Interval::Contains(double p) const {
   // Works for empty, full, and singleton intervals.
-  ABSL_DCHECK_LE(fabs(p), M_PI);
+  ABSL_DCHECK(IsValidPoint(p));
```
jmr: error message wasn't great before, but is worse now.  `ABSL_DCHECK() << p;`

src/python/s1interval_bindings.cc
```diff
@@ -0,0 +1,146 @@
...
+#include <sstream>
+
+#include "s2/s1interval.h"
+
+namespace py = pybind11;
+
+namespace {
+
+void MaybeThrowInvalidPoint(double p) {
+  if (!S1Interval::IsValidPoint(p)) throw py::value_error("Invalid S1 point");
```
jmr: more info here, like value of p

jmr (changes requested):

src/python/s2point_test.py
```diff
@@ -1,12 +1,14 @@
 """Tests for S2Point pybind11 bindings."""
 
+import math
 import unittest
 import s2geometry_pybind as s2
 
-
```
jmr, line 6: this blank line got deleted, but should be there:
https://google.github.io/styleguide/pyguide.html#35-blank-lines

src/python/s2point_test.py
```diff
@@ -19,6 +21,39 @@ def test_constructor_with_coordinates(self):
...
+            s2.S2Point((1.0, 2.0, 3.0, 4.0))
+
+    # Properties
+
+    def test_data(self):
+        p = s2.S2Point(1.0, 2.0, 3.0)
+        coords = p.data()
+        self.assertEqual(coords, (1.0, 2.0, 3.0))
+
+    # Vector operations.
```
jmr: nit: no trailing period for consistency

src/python/s1interval_bindings.cc
jmr: Hmm, so we actually throw a C++ exception here.  Is there a way to make this work without C++ exceptions, like returning an error-like type that then gets converted to an exception in Python?

src/s2/s1interval.cc
deustis: Done

src/python/s1interval_bindings.cc
deustis: I'm pretty sure this is the recommended approach:
https://pybind11.readthedocs.io/en/stable/advanced/exceptions.html

I haven't found any other ways to make pybind raise a Python exception without throwing a C++ exception.

deustis: Done

src/python/s2point_test.py
deustis: Fixed

src/python/s2point_test.py
deustis: Done

src/python/s1interval_bindings.cc
```diff
@@ -0,0 +1,146 @@
...
+#include <sstream>
+
+#include "s2/s1interval.h"
+
+namespace py = pybind11;
+
+namespace {
+
+void MaybeThrowInvalidPoint(double p) {
+  if (!S1Interval::IsValidPoint(p)) throw py::value_error("Invalid S1 point: " + std::to_string(p));
```
jmr: Line length.

src/python/s1interval_bindings.cc
jmr: Add a comment to ehe BUILD and/or README that exceptions need to be enabled.

src/python/s1interval_bindings.cc
deustis: Fixed

src/python/s1interval_bindings.cc
deustis: Done, added to README in the Invalid Values section

PR #545, opened 2026-02-19
MSVC Fix

James-E-A:
For some reason, Microsoft's compiler doesn't seem to infer these overloads correctly. This patch allows successful compilation of **v0.13.1**.

### DIFF
--- a/src/s2/util/math/exactfloat/exactfloat.cc
+++ b/src/s2/util/math/exactfloat/exactfloat.cc
@@ -151,8 +151,8 @@ ExactFloat::ExactFloat(double v) {
     // by the number of mantissa bits in a double (53, including the leading
     // "1") then the result is always an integer.
     int exp;
-    double f = frexp(fabs(v), &exp);
-    uint64_t m = static_cast<uint64_t>(ldexp(f, kDoubleMantissaBits));
+    double f = std::frexp(std::fabs(v), &exp);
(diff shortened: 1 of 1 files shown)


src/s2/util/math/exactfloat/exactfloat.cc
```diff
@@ -25,11 +25,19 @@
...
 #include "absl/strings/str_format.h"
 
+#ifndef OPENSSL_IS_BORINGSSL
+#include "absl/container/fixed_array.h"  // IWYU pragma: keep
+#include "absl/numeric/bits.h"  // IWYU pragma: keep
+#endif
+
 namespace exactfloat {
 
 using std::max;
```
jmr, line 43: `using std::abs;` etc would be more consistent with the current coding style.  Does that work?

jmr:
Does https://github.com/google/s2geometry/pull/555 work for you?

James-E-A:
I've been very busy with other projects recently; my feelings won't be hurt at all if someone beats me to the punch on figuring out how to translate this patch from `v0.13.1` — which is still the latest release! Is there any possibility of doing a *.2* bump with just this patch? — to whatever's going on on `master`. If not, I expect to have this checked out again by mid-April.

And RE `using std::abs;` I am also *not* familiar with this codebase or C++ generally; if there's alleged a prettier alternative fix, I have nothing to say in defense of my own original approach.

PR #546, opened 2026-02-22
pybind: Add R2 bindings

deustis:
Add pybind11 bindings for R1Interval, R2Point, and R2Rect

### DIFF
--- a/src/python/README.md
+++ b/src/python/README.md
@@ -57,8 +57,7 @@ The Python bindings follow the C++ API closely but with Pythonic conventions:
 - Standard Python operators work as expected: `+`, `-`, `*`, `==`, `!=`, `<`, `>` (for C++ classes that implement those operators)
 
 **Method Overloads:**
-- Some methods accept more than one type for the same argument. For example, `R2Rect.contains()` accepts either an `R2Point` or another `R2Rect`.
-- When a C++ class has overloads with significantly different behavior, the Python bindings use distinct method names instead. For example, `R2Rect` provides `vertex(k)` for CCW vertex index and `vertex_ij(i, j)` for axis-direction vertex access.
+- Instead of Python overloads, C++ overloaded methods are exposed under distinct names. For different argument types, the short name is used for the same-type argument and a longer name for other types (e.g., `R2Rect.contains(other: R2Rect)` and `R2Rect.contains_point(p: R2Point)`). For overloads with significantly different behavior, descriptive names are chosen (e.g., `R2Rect.vertex(k)` for CCW vertex index and `R2Rect.vertex_ij(i, j)` for axis-direction vertex access).
 
 **Default Values:**
 - Some methods have default parameter values (e.g., `approx_equals(other, max_error=1e-15)`). These defaults are re-defined in the Python bindings and are expected to match the C++ defaults.
--- a/src/python/r2rect_bindings.cc
+++ b/src/python/r2rect_bindings.cc
@@ -104,11 +104,11 @@ void bind_r2rect(py::module& m) {
       .def("size", &R2Rect::GetSize,
            "Return the width and height of the rectangle.\n\n"
            "Empty rectangles have a negative width and height.")
-      .def("contains", py::overload_cast<const R2Point&>(
+      .def("contains_point", py::overload_cast<const R2Point&>(
                &R2Rect::Contains, py::const_),
            py::arg("p"),
            "Return true if the rectangle contains the given point")
-      .def("interior_contains", py::overload_cast<const R2Point&>(
+      .def("interior_contains_point", py::overload_cast<const R2Point&>(
                &R2Rect::InteriorContains, py::const_),
            py::arg("p"),
            "Return true if the interior of the rectangle contains the given "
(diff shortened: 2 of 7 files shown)


src/python/r2rect_bindings.cc
```diff
@@ -0,0 +1,177 @@
...
+namespace {
+
+void MaybeThrowInvalidRect(const R2Point& lo, const R2Point& hi) {
+  // Both x and y intervals must be either both empty or both non-empty.
+  bool x_empty = lo.x() > hi.x();
+  bool y_empty = lo.y() > hi.y();
+  if (x_empty != y_empty) {
+    throw py::value_error(
+        "Invalid R2Rect: x and y intervals must be both empty or both "
+        "non-empty");
```
jmr: more info here

src/python/r2rect_bindings.cc
```diff
@@ -0,0 +1,177 @@
...
+
+      // Geometric operations
+      .def("vertex",
+           py::overload_cast<int>(&R2Rect::GetVertex, py::const_),
+           py::arg("k"),
+           "Return the k-th vertex of the rectangle (k = 0,1,2,3) in CCW "
+           "order.\n\n"
+           "Vertex 0 is in the lower-left corner. The argument is reduced "
+           "modulo 4.")
+      .def("vertex_ij",
```
jmr, line 95: A single overloaded `vertex` might make sense. I'm not sure and don't have strong feelings.

deustis:
> I'm also not sure about the "expose all the things" approach. We've never exposed R2* in the old SWIG or new google3 bindings and no one complained. Maybe you want to do a useful subset / the SWIG subset first.

I believe `R1Interval` is exposed in SWIG, but you're right that the others are not. It looks like methods on S2CellID like `R2Point GetCenterUV() const;` are also not exposed. I was trying to be a bit more comprehensive but I'll defer to your judgement...

How about this?: Let's try to land this PR and then I'll move on to some more useful classes like `S2LatLng`, `S2CellId`, and `S2Cell`. If I hit more of these missing primitives we can have a discussion before exposing them.

src/python/r2rect_bindings.cc
deustis: I left a note about this here:
https://github.com/google/s2geometry/pull/546/changes#diff-68b4b8a36a193cbacbd9e072b0026059a3554ed10ee9779d07856a8110628193R61

Python overloads are generally frowned on but if you prefer to match the C++ behavior I could get behind that too. Or we could drop this method as it is pretty obscure.

src/python/r2rect_bindings.cc
deustis: Done

deustis:
@jwr, please see above and let me know how you'd like to proceed?:
https://github.com/google/s2geometry/pull/546#issuecomment-4006468331

src/python/BUILD.bazel
```diff
@@ -26,11 +26,38 @@ pybind_extension(
...
         ":s1interval_bindings",
         ":s2point_bindings",
     ],
 )
 
+pybind_library(
+    name = "r1interval_bindings",
+    srcs = ["r1interval_bindings.cc"],
+    deps = [
+        "//:s2",
```
jmr, line 41: I was going to say use the fine-grained dep, but we don't have that. Yet, so ok.

src/python/README.md
```diff
@@ -56,6 +56,13 @@ The Python bindings follow the C++ API closely but with Pythonic conventions:
 **Operators:**
 - Standard Python operators work as expected: `+`, `-`, `*`, `==`, `!=`, `<`, `>` (for C++ classes that implement those operators)
 
+**Method Overloads:**
+- Some methods accept more than one type for the same argument. For example, `R2Rect.contains()` accepts either an `R2Point` or another `R2Rect`.
```
jmr: What we do in the pyclif bindings is have the "short" name for the same type and a longer name for other types, so `R2Rect.contains(R2Rect)` and `R2Rect.contains_point(R2Point)`.

src/python/README.md
```diff
@@ -56,6 +56,13 @@ The Python bindings follow the C++ API closely but with Pythonic conventions:
...
+- When a C++ class has overloads with significantly different behavior, the Python bindings use distinct method names instead. For example, `R2Rect` provides `vertex(k)` for CCW vertex index and `vertex_ij(i, j)` for axis-direction vertex access.
```
jmr: Not for this PR, but if we consider these "significantly different", we should rename them in C++, too. I was thinking of them as pretty much the same, but it could make sense to rename in C++.

jmr:
One thing would be to support `__hash__`, but let's just finish this one and come back to that later. #553

src/python/README.md
deustis: > What we do in the pyclif bindings is have the "short" name for the same type and a longer name for other types, so `R2Rect.contains(R2Rect)` and `R2Rect.contains_point(R2Point)`.

Sounds good. I have made this change. Also updated one overload from s1interval.

deustis:
@jmr, friendly nudge, let me know you'd like to see any other changes on this?

jmr:
Sorry about the delay. Thanks for the contribution!

PR #562, opened 2026-04-07
pybind: Add S1Angle bindings

deustis:
Add pybind11 bindings for S1Angle.

Also add S1Angle::IsNormalizedAngle() to s1angle.h for pre-validating
E5/E6/E7 conversions.

Supports downstream PRs for S2LatLng (deustis/s2geometry#2) and
S2CellId (deustis/s2geometry#3).

### DIFF
--- a/src/python/module.cc
+++ b/src/python/module.cc
@@ -6,6 +6,7 @@ namespace py = pybind11;
 void bind_r1interval(py::module& m);
 void bind_r2point(py::module& m);
 void bind_r2rect(py::module& m);
+void bind_s1angle(py::module& m);
 void bind_s1interval(py::module& m);
 void bind_s2point(py::module& m);
 
@@ -18,4 +19,5 @@ PYBIND11_MODULE(s2geometry_bindings, m) {
   bind_r2rect(m);
   bind_s1interval(m);
   bind_s2point(m);
+  bind_s1angle(m);
 }
--- a/src/python/module.cc
+++ b/src/python/module.cc
@@ -13,11 +13,19 @@ void bind_s2point(py::module& m);
 PYBIND11_MODULE(s2geometry_bindings, m) {
   m.doc() = "S2 Geometry Python bindings using pybind11";
 
-  // Bind core geometry classes in dependency order
+  // Bind core geometry classes in dependency order.
+  // Each class must be registered before classes that use it as a
+  // parameter or return type.
+
+  // No dependencies
   bind_r1interval(m);
   bind_r2point(m);
-  bind_r2rect(m);
   bind_s1interval(m);
   bind_s2point(m);
+
+  // Deps: r1interval, r2point
+  bind_r2rect(m);
+
+  // Deps: s2point
   bind_s1angle(m);
 }
--- a/src/python/s1angle_bindings.cc
+++ b/src/python/s1angle_bindings.cc
@@ -12,7 +12,7 @@ namespace py = pybind11;
 namespace {
 
 void MaybeThrowNotNormalized(const S1Angle& angle) {
-  if (!angle.IsNormalizedAngle()) {
+  if (!angle.IsNormalized()) {
     throw py::value_error("Angle " + std::to_string(angle.degrees()) +
                           " degrees is not in the normalized range (-180, 180]");
   }
@@ -96,15 +96,10 @@ void bind_s1angle(py::module& m) {
            "Return the E7 representation (degrees * 1e7, rounded).\n\n"
            "The angle must be in the normalized range (-180, 180] degrees.\n"
            "Raises ValueError if out of range.")
-      .def("abs", &S1Angle::abs, "Return the absolute value of the angle")
+      .def("__abs__", &S1Angle::abs,
+           "Return the absolute value of the angle")
       .def("normalized", &S1Angle::Normalized,
            "Return the angle normalized to the range (-180, 180] degrees")
-      .def("sin_cos", [](const S1Angle& self) {
-               auto sc = self.SinCos();
-               return py::make_tuple(sc.sin, sc.cos);
-           },
-           "Return (sin, cos) of the angle as a tuple.\n\n"
-           "This may be more efficient than calling sin and cos separately.")
 
       // Operators
       .def(py::self == py::self, "Return true if angles are exactly equal")
--- a/src/python/s1angle_bindings.cc
+++ b/src/python/s1angle_bindings.cc
@@ -1,0 +1,158 @@
+#include <pybind11/pybind11.h>
+#include <pybind11/operators.h>
+
+#include <sstream>
+#include <string>
+
+#include "s2/s1angle.h"
+#include "s2/s2point.h"
+
+namespace py = pybind11;
+
+namespace {
+
+void MaybeThrowNotNormalized(const S1Angle& angle) {
+  if (!angle.IsNormalizedAngle()) {
+    throw py::value_error("Angle " + std::to_string(angle.degrees()) +
+                          " degrees is not in the normalized range (-180, 180]");
+  }
+}
+
+}  // namespace
+
+void bind_s1angle(py::module& m) {
+  py::class_<S1Angle>(m, "S1Angle",
+      "Represents a one-dimensional angle.\n\n"
+      "The internal representation is a double-precision value in radians,\n"
+      "so conversion to and from radians is exact. Conversions between\n"
+      "degrees and radians are not always exact due to floating-point\n"
+      "arithmetic (e.g. from_degrees(60).degrees() != 60). Use e5/e6/e7\n"
+      "representations for exact discrete comparisons.\n\n"
+      "See s2/s1angle.h for comprehensive documentation including exact\n"
+      "conversion guarantees and edge cases.")
+      // Constructors
+      .def(py::init<>(), "Default constructor creates a zero angle")
+      .def(py::init<const S2Point&, const S2Point&>(),
+           py::arg("x"), py::arg("y"),
+           "Construct the angle between two points.\n\n"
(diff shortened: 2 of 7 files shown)


deustis:
@jmr, ready for you review. Not sure why the CI is failing

src/s2/s1angle.h
```diff
@@ -130,6 +130,10 @@ class S1Angle {
   int32_t e6() const;
   int32_t e7() const;
 
+  // Returns true if the angle is in the normalized range (-180, 180]
+  // degrees, which is the valid input range for e5(), e6(), and e7().
+  bool IsNormalizedAngle() const;
```
jmr: I think "Angle" is implied based on the type.  Why not call this `IsNormalized` and move it to just before the `Normalized()` declaration/definition?

src/python/s1angle_bindings.cc
```diff
@@ -0,0 +1,158 @@
...
+               MaybeThrowNotNormalized(self);
+               return self.e7();
+           },
+           "Return the E7 representation (degrees * 1e7, rounded).\n\n"
+           "The angle must be in the normalized range (-180, 180] degrees.\n"
+           "Raises ValueError if out of range.")
+      .def("abs", &S1Angle::abs, "Return the absolute value of the angle")
+      .def("normalized", &S1Angle::Normalized,
+           "Return the angle normalized to the range (-180, 180] degrees")
+      .def("sin_cos", [](const S1Angle& self) {
```
jmr: It's Python.  Are we really worried about this?

src/python/s1angle_bindings.cc
```diff
@@ -0,0 +1,158 @@
...
+           "The angle must be in the normalized range (-180, 180] degrees.\n"
+           "Raises ValueError if out of range.")
+      .def("e7", [](const S1Angle& self) {
+               MaybeThrowNotNormalized(self);
+               return self.e7();
+           },
+           "Return the E7 representation (degrees * 1e7, rounded).\n\n"
+           "The angle must be in the normalized range (-180, 180] degrees.\n"
+           "Raises ValueError if out of range.")
+      .def("abs", &S1Angle::abs, "Return the absolute value of the angle")
```
jmr: Support `__abs__`?

jmr (changes requested):

src/s2/s1angle.h
deustis: Done.

src/python/s1angle_bindings.cc
deustis: Dropped `sin_cos()`

src/python/s1angle_bindings.cc
deustis: Done.

deustis:
@jmr, I believe I've addressed your comments!

src/python/s1angle_test.py
```diff
@@ -0,0 +1,274 @@
...
+    def test_in_place_scalar_division(self):
+        a = s2.S1Angle.from_degrees(90.0)
+        a /= 2.0
+        self.assertAlmostEqual(a.degrees(), 45.0)
+
+    # String representation
+
+    def test_repr(self):
+        a = s2.S1Angle.from_degrees(45.0)
+        self.assertEqual(repr(a), "S1Angle(45.0000000)")
```
jmr, line 247: hmm. add `assertEqual(a, S1Angle(45.0000000))` here. I don't think we have a float constructor.  Use `from_degrees` and ideally control the number of trailing 0s.

src/python/s1angle_test.py
```diff
@@ -0,0 +1,274 @@
...
+        # Exact conversions per the C++ header documentation.
+        self.assertEqual(s2.S1Angle.from_degrees(180.0).radians(), math.pi)
+        for k in range(9):
+            self.assertEqual(
+                s2.S1Angle.from_degrees(45 * k).radians(),
+                k * math.pi / 4)
+
+    def test_e5(self):
+        a = s2.S1Angle.from_degrees(45.0)
+        self.assertEqual(a.e5(), 4500000)
```
jmr, line 86: I think all these `eN` / `radians` / `degrees` should be properties.

src/python/s1angle_bindings.cc
```diff
@@ -0,0 +1,153 @@
...
+      .def("e7", [](const S1Angle& self) {
+               MaybeThrowNotNormalized(self);
+               return self.e7();
+           },
+           "Return the E7 representation (degrees * 1e7, rounded).\n\n"
+           "The angle must be in the normalized range (-180, 180] degrees.\n"
+           "Raises ValueError if out of range.")
+      .def("__abs__", &S1Angle::abs,
+           "Return the absolute value of the angle")
+      .def("normalized", &S1Angle::Normalized,
```
jmr, line 101: might be worthwhile to expose Normalize()

jmr (changes requested):
Sorry, I noticed a few more things.

src/python/module.cc
```diff
@@ -18,4 +19,5 @@ PYBIND11_MODULE(s2geometry_bindings, m) {
   bind_r2rect(m);
   bind_s1interval(m);
   bind_s2point(m);
+  bind_s1angle(m);
```
jmr, line 22: This says "dependency order", but eithe make it alphabetic if this has no deps, or comment what the dep is.

src/python/s1angle_bindings.cc
```diff
@@ -0,0 +1,153 @@
...
+#include "s2/s1angle.h"
+#include "s2/s2point.h"
+
+namespace py = pybind11;
+
+namespace {
+
+void MaybeThrowNotNormalized(const S1Angle& angle) {
+  if (!angle.IsNormalized()) {
+    throw py::value_error("Angle " + std::to_string(angle.degrees()) +
```
jmr, line 16: We should avoid `std::to_string` since it uses locale.  (I missed this in a previous PR.) Use `absl::StrCat` here instead.

src/python/s1angle_bindings.cc
jmr: `IsNormalized`, too

jmr (changes requested):

jmr:
> Supports downstream PRs for S2LatLng (https://github.com/deustis/s2geometry/pull/2) and
S2CellId (https://github.com/deustis/s2geometry/pull/3).

GH now supports stacked PRs. I'm not sure how well they work, or if they would be helpful in this case.

src/python/s1angle_test.py
```diff
@@ -0,0 +1,274 @@
...
+
+    def test_repr_zero(self):
+        a = s2.S1Angle.zero()
+        self.assertEqual(repr(a), "S1Angle(0.0000000)")
+
+    # Module-level functions
+
+    def test_sin(self):
+        a = s2.S1Angle.from_degrees(90.0)
+        self.assertAlmostEqual(s2.sin(a), 1.0)
```
jmr, line 261: Is this is going to be a problem when we have `S1ChordAngle.sin`?

Will `s2.sin` have type cases to handle both types?

Maybe better to go with `a.sin()`.  Writing `s2.sin(a)` doesn't seem that natural anyway.

src/python/s1angle_test.py
deustis: Not quite sure if this addresses your comment, but added test variations that show the input precision doesn't affect the output precision (since that's controlled by the C++ implementation)

src/python/s1angle_test.py
deustis: `radians`, `degrees`, `e5`, `e6`, `e7` are now properties. Updated the README property rule to cover simple accessors including trivial unit conversions.

src/python/s1angle_bindings.cc
deustis: Exposed both `is_normalized()` and `normalize()`

src/python/module.cc
deustis: Reorganized into dependency tranches with per-line comments (e.g. `// Deps: s2point`).

src/python/s1angle_bindings.cc
deustis: Done — switched to `absl::StrCat` in both `s1angle_bindings.cc` and `s1interval_bindings.cc`.

src/python/s1angle_test.py
deustis: Moved `sin`/`cos`/`tan` from module-level functions to instance methods (`a.sin()`, `a.cos()`, `a.tan()`). This avoids the S1ChordAngle collision and reads more naturally.

deustis:
> > Supports downstream PRs for S2LatLng ([deustis#2](https://github.com/deustis/s2geometry/pull/2)) and
> > S2CellId ([deustis#3](https://github.com/deustis/s2geometry/pull/3)).
> 
> GH now supports stacked PRs. I'm not sure how well they work, or if they would be helpful in this case.

Thanks... we'll just take these one at a time

deustis:
@jmr, PTAL!

src/python/s1angle_test.py
jmr: I think we shouldn't be using the C++ operator here, but let's not block this PR any longer.

https://github.com/google/s2geometry/issues/578

src/python/s1interval_bindings.cc
```diff
@@ -1,6 +1,8 @@
 #include <pybind11/pybind11.h>
 #include <pybind11/operators.h>
 
+#include <sstream>
```
jmr: We don't need this now.

jmr (approved):
Good except for the extra `#include`.

src/python/s1interval_bindings.cc
deustis: Removed

PR #569, opened 2026-04-18
Fix bignum.cc build failure on GCC 8 (x86_64)

### LINKED ISSUE #568: bignum.cc: _addcarry_u64/_subborrow_u64 build failure on GCC 8 (x86_64)
src/s2/util/math/exactfloat/bignum.cc uses _addcarry_u64 and _subborrow_u64 intrinsics guarded by #ifdef __x86_64__. 
However, these intrinsics require the ADX instruction set extension, which GCC < 9 does not enable by default on x86_64.

This causes a build failure on GCC 8.5.0 

Proposed fix
Replace #ifdef __x86_64__ with a check that also verifies ADX support:

#if defined(__x86_64__) && (defined(__clang__) || (defined(__GNUC__) && __GNUC__ >= 9) || defined(__ADX__))
This allows the intrinsics on:

Clang (all versions support them on x86_64)
GCC >= 9 (enables ADX by default)
Any compiler with -madx flag (__ADX__ is defined)
And falls through to the portable absl::uint128 path on older GCC, with no functional change.

Environment
GCC 8.5.0 on Rocky Linux 8 / RHEL 8 (x86_64)

akhanzode:
## Description

The `_addcarry_u64` and `_subborrow_u64` intrinsics in `bignum.cc` require the ADX instruction set extension, which GCC < 9 does not enable by default on x86_64. The existing `#ifdef __x86_64__` guard is too broad, causing build failures on GCC 8.5.0 (the default compiler on Rocky Linux 8 / RHEL 8).

```
bignum.cc:299:7: error: '_addcarry_u64' was not declared in this scope
bignum.cc:316:13: error: '_subborrow_u64' was not declared in this scope
```

## Fix

Replace `#ifdef __x86_64__` with a stricter check:

```cpp
#if defined(__x86_64__) && (defined(__clang__) || (defined(__GNUC__) && __GNUC__ >= 9) || defined(__ADX__))
```

This allows the intrinsics on Clang (all versions), GCC >= 9, or any compiler with `-madx`. On older GCC, falls through to the existing portable `absl::uint128` arithmetic path with no functional change.

Fixes #568

### DIFF
--- a/src/s2/util/math/exactfloat/bignum.cc
+++ b/src/s2/util/math/exactfloat/bignum.cc
@@ -15,7 +15,7 @@
 
 #include "s2/util/math/exactfloat/bignum.h"
 
-#ifdef __x86_64__
+#if defined(__x86_64__) && (defined(__clang__) || (defined(__GNUC__) && __GNUC__ >= 9) || defined(__ADX__))
 #include <immintrin.h>
 #endif
 
@@ -292,7 +292,7 @@ inline Bigit AddBigit(Bigit a, Bigit b, Bigit* absl_nonnull carry) {
 // Godbolt link for comparison:
 //   https://godbolt.org/z/cGnnfMbMn  (no intrinsics)
 //   https://godbolt.org/z/jnM1Y3Tjs  (intrinsics)
-#ifdef __x86_64__
+#if defined(__x86_64__) && (defined(__clang__) || (defined(__GNUC__) && __GNUC__ >= 9) || defined(__ADX__))
   static_assert(sizeof(Bigit) == sizeof(unsigned long long));
   Bigit out;
   *carry =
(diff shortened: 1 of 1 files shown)


jmr:
https://github.com/google/s2geometry/issues/568#issuecomment-4275540304

jmr:
You can replace `__x86_64__` with `__ADX__` if that's really all we need. If it happens to fix RHEL 8, great, but no guarantees with that.

https://github.com/google/s2geometry/issues/568#issuecomment-4275540304

jmr:
I think `__ADX__` is the wrong macro here. It indicates support for the `ADCX`/`ADOX` instructions (Intel ADX extension), but `_addcarry_u64` and `_subborrow_u64` map to plain `ADC` and `SBB`, which are base x86 instructions — no ADX required.

PR #577, opened 2026-04-21
python: Use Limited API / Stable ABI

jmr:
Introduce a new CMake option, `WITH_PYTHON_LIMITED_API`, to optionally build the SWIG Python bindings using the Python Limited API (Stable ABI). This allows the generated bindings to be compatible across multiple Python 3.10+ minor versions without recompilation.

When `WITH_PYTHON_LIMITED_API` is enabled:
- The build requires SWIG >= 4.2, and CMake >= 3.26.
- The `Py_LIMITED_API` macro is defined for the compiler.

When disabled (the default), the traditional Python C API is used, preserving compatibility with older toolchains (SWIG 4.0).

https://docs.python.org/3/c-api/stable.html

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,7 +49,8 @@ jobs:
       - run: mkdir build
       - run: >
           cmake
-          -DWITH_PYTHON=yes
+          -DWITH_PYTHON=ON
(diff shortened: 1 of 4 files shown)


jmr:
Ok. I flipped the default to off, since it might confuse people and is probably only useful for the wheell.

PR #581, opened 2026-04-24
pybind: Add S2LatLng bindings

deustis:
Add pybind11 bindings for S2LatLng.

Supports downstream PR for S2CellId.

### DIFF
--- a/src/python/BUILD.bazel
+++ b/src/python/BUILD.bazel
@@ -32,6 +32,7 @@ pybind_extension(
         ":s1angle_bindings",
         ":s1interval_bindings",
         ":s2point_bindings",
+        ":s2latlng_bindings",
     ],
 )
 
@@ -84,6 +85,14 @@ pybind_library(
     ],
 )
 
+pybind_library(
+    name = "s2latlng_bindings",
+    srcs = ["s2latlng_bindings.cc"],
+    deps = [
+        "//:s2",
+    ],
+)
+
 # ========================================
 # Python Tests
 # ========================================
@@ -121,5 +130,11 @@ py_test(
 py_test(
     name = "s2point_test",
     srcs = ["s2point_test.py"],
+    deps = [":s2geometry_pybind"],
+)
+
+py_test(
+    name = "s2latlng_test",
+    srcs = ["s2latlng_test.py"],
     deps = [":s2geometry_pybind"],
 )
--- a/src/python/BUILD.bazel
+++ b/src/python/BUILD.bazel
@@ -90,6 +90,7 @@ pybind_library(
     srcs = ["s2latlng_bindings.cc"],
     deps = [
         "//:s2",
+        "@abseil-cpp//absl/strings",
     ],
 )
 
--- a/src/python/README.md
+++ b/src/python/README.md
@@ -43,6 +43,8 @@ The Python bindings follow the C++ API closely but with Pythonic conventions:
 **Invalid Values:**
 - Invalid inputs to constructions or functions raises `ValueError`.
 - Example: `S1Interval(0.0, 4.0)` raises `ValueError` because `4.0 > π`.
+- Some classes change the behavior of operations to guarantee validity.
+- Example: `S2LatLng` arithmetic operators automatically clamp the result to a valid range whereas in C++, `S2LatLng` these operators can produce invalid states.
 - Note: In C++, these conditions trigger `ABSL_DCHECK` assertions. The bindings prevent these assertions from firing by pre-validating inputs.
 - Note: Python bindings check for invalid inputs and throw C++ exceptions which are caught by
   pybind and converted to Python exceptions. Exceptions are normally prohibited by the C++
--- a/src/python/README.md
+++ b/src/python/README.md
@@ -44,7 +44,7 @@ The Python bindings follow the C++ API closely but with Pythonic conventions:
 - Invalid inputs to constructions or functions raises `ValueError`.
 - Example: `S1Interval(0.0, 4.0)` raises `ValueError` because `4.0 > π`.
 - Some classes change the behavior of operations to guarantee validity.
-- Example: `S2LatLng` arithmetic operators automatically clamp the result to a valid range whereas in C++, `S2LatLng` these operators can produce invalid states.
+- Example: `S2LatLng` arithmetic operators automatically clamp the result to a valid range whereas in C++ these operators can produce invalid states.
 - Note: In C++, these conditions trigger `ABSL_DCHECK` assertions. The bindings prevent these assertions from firing by pre-validating inputs.
 - Note: Python bindings check for invalid inputs and throw C++ exceptions which are caught by
   pybind and converted to Python exceptions. Exceptions are normally prohibited by the C++
--- a/src/python/module.cc
+++ b/src/python/module.cc
@@ -29,5 +29,7 @@ PYBIND11_MODULE(s2geometry_bindings, m) {
 
   // Deps: s2point
   bind_s1angle(m);
+
+  // Deps: s1angle, s2point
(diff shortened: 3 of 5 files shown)


src/python/s2latlng_bindings.cc
```diff
@@ -0,0 +1,187 @@
...
+           "Construct from latitude and longitude angles.\n\n"
+           "Raises ValueError if latitude is not in [-90, 90] or\n"
+           "longitude is not in [-180, 180] degrees.")
+      .def(py::init([](const S2Point& p) {
+               S2LatLng ll(p);
+               MaybeThrowNotValid(ll);
+               return ll;
+           }),
+           py::arg("point"),
+           "Convert a direction vector to an S2LatLng.\n\n"
```
deustis, line 59: Where does this term "direction vector" come from? Consider switching to "S2Point vector" here and elsewhere.

src/python/s2latlng_bindings.cc
```diff
@@ -0,0 +1,187 @@
...
+           "The vector does not need to be unit length.\n"
+           "Raises ValueError if the point has non-finite coordinates.")
+
+      // Factory methods
+      .def_static("from_radians", [](double lat_radians, double lng_radians) {
```
deustis, line 64: Should we provide a variant like `from_radians_normalized` that does the clamping on latitude and normalization on longitude? Do the same for other `from_` constructors? Or alternatively stand-alone static functions like `clamp_latitude_radians` and `normalize_longitude_radians`?

src/python/s2latlng_bindings.cc
```diff
@@ -0,0 +1,187 @@
...
+      .def_static("latitude", &S2LatLng::Latitude, py::arg("point"),
+                  "Return the latitude of a direction vector")
+      .def_static("longitude", &S2LatLng::Longitude, py::arg("point"),
+                  "Return the longitude of a direction vector")
+
+      // Properties
+      .def_property_readonly("lat", &S2LatLng::lat, "The latitude as an S1Angle")
+      .def_property_readonly("lng", &S2LatLng::lng,
+                             "The longitude as an S1Angle")
+      .def("coords", [](const S2LatLng& self) {
```
deustis: The C++ returns an `R2Point`, shouldn't we do that as well? And make this a readonly property

src/python/s2latlng_bindings.cc
```diff
@@ -0,0 +1,187 @@
...
+      }, "Return the (lat, lng) coordinates in radians as a tuple")
+
+      // Geometric operations
+      .def("to_point", &S2LatLng::ToPoint,
+           "Convert to the equivalent unit-length S2Point")
+      .def("get_distance", &S2LatLng::GetDistance,
+           py::arg("other"),
+           "Return the surface distance to another S2LatLng.\n\n"
+           "Uses the Haversine formula.")
+      .def("to_string_in_degrees", &S2LatLng::ToStringInDegrees,
```
deustis: Move to string representation section

src/python/s2latlng_bindings.cc
deustis: The term comes from upstream — s2/s2latlng.h:60 describes this constructor as "Convert a direction vector (not necessarily unit length) to an S2LatLng", and the term is used consistently across s2coords.h, s2cell.h, and s2cell_id.h. Keeping the bindings aligned with the upstream docstring.

src/python/s2latlng_bindings.cc
deustis: Added `from_radians_normalized` for now. Can extend to degrees/E5/E6/E7 variants in a follow-up if needed.

src/python/s2latlng_bindings.cc
deustis: Done — `coords` is now a readonly property returning R2Point.

src/python/s2latlng_bindings.cc
deustis: Done.

deustis:
@jmr, next installment here. Getting to the fun/useful stuff.

src/python/s2latlng_bindings.cc
```diff
@@ -0,0 +1,197 @@
...
+      // Factory methods
+      .def_static("from_radians", [](double lat_radians, double lng_radians) {
+               S2LatLng ll = S2LatLng::FromRadians(lat_radians, lng_radians);
+               MaybeThrowNotValid(ll);
+               return ll;
+           },
+           py::arg("lat_radians"), py::arg("lng_radians"),
+           "Construct from latitude and longitude in radians.\n\n"
+           "Raises ValueError if out of range.")
+      .def_static("from_radians_normalized",
```
deustis: @jmr, would like to get your take here. Should all of the constructors just normalize by default by clamping latitude and normalizing longitude? Or is it better to be strict and raise value errors? I tentatively went with the latter and added this one variant to give users a way to normalize but open to changing this.

src/python/README.md
```diff
@@ -43,6 +43,8 @@ The Python bindings follow the C++ API closely but with Pythonic conventions:
 **Invalid Values:**
 - Invalid inputs to constructions or functions raises `ValueError`.
 - Example: `S1Interval(0.0, 4.0)` raises `ValueError` because `4.0 > π`.
+- Some classes change the behavior of operations to guarantee validity.
+- Example: `S2LatLng` arithmetic operators automatically clamp the result to a valid range whereas in C++, `S2LatLng` these operators can produce invalid states.
```
jmr: second `S2LatLng` seems to be stray

src/python/BUILD.bazel
```diff
@@ -84,6 +85,14 @@ pybind_library(
     ],
 )
 
+pybind_library(
+    name = "s2latlng_bindings",
+    srcs = ["s2latlng_bindings.cc"],
+    deps = [
+        "//:s2",
```
jmr, line 92: missing absl/strings BUILD dep

src/python/s2latlng_bindings.cc
jmr: https://peps.python.org/pep-0020/#:~:text=Errors%20should%20never%20pass%20silently.

I think strict is good.

jmr: `from_radians_normalized` sounds too much like "from normalized radians" to me.

How about either `from_unnormalized_radians` or `normalized_from_radians`?

src/python/README.md
deustis: Done. Dropped the stray second `S2LatLng`.

src/python/BUILD.bazel
deustis: Done. Added `@abseil-cpp//absl/strings` to the `s2latlng_bindings` deps.

src/python/s2latlng_bindings.cc
deustis: Acknowledged, keeping strict.

deustis: Done. Renamed to `normalized_from_radians`.

deustis:
@jmr, let me know if you have any other suggestions on this PR? I'm not sure why the cmake tests are failing...

src/python/module.cc
```diff
@@ -9,6 +9,7 @@ void bind_r2rect(py::module& m);
 void bind_s1angle(py::module& m);
 void bind_s1interval(py::module& m);
 void bind_s2point(py::module& m);
+void bind_s2latlng(py::module& m);
```
jmr, line 12: let's keep these sorted alphabetically.  You can add `// keep-sorted start` and `// keep-sorted end` if you want.

jmr (approved):
We can do the sorting change separately.

PR #593, opened 2026-05-04
pybind: Add S2CellId bindings

deustis:
Add pybind11 bindings for S2CellId.

Iteration-like methods (e.g. next(), prev(), Begin(), End()) are not included.

Also updates the README with new binding file organization sections (Constants, Traversal).

(Part of a series addressing https://github.com/google/s2geometry/pull/524)

### DIFF
--- a/src/python/README.md
+++ b/src/python/README.md
@@ -76,6 +76,10 @@ The Python bindings follow the C++ API closely but with Pythonic conventions:
 - C++ functions that accept or return a vector object use a Python tuple (of length matching the vector dimension)
 - Array indexing operators (e.g., `point[0]`) are not currently supported
 
+**Iterators:**
+- Some classes support `__iter__` for traversal (e.g., `for c in S2CellId.cells(level=5):`) or expose methods that return iterables.
+- Iterables support forward iteration, `len()`, indexing, `in`, and `reversed()` unless otherwise noted.
+
 **Serialization:**
 - The C++ Encoder/Decoder serialization functions are not currently supported
 
@@ -145,10 +149,12 @@ Use the following sections to organize functions within the bindings files and t
 
 1. **Constructors** - Default constructors and constructors with parameters
 2. **Factory methods** - Static factory methods (e.g., `from_degrees`, `from_radians`, `zero`, `invalid`)
-3. **Properties** - Mutable and read-only properties (e.g., coordinate accessors like `x`, `y`, `lo`, `hi`)
-4. **Predicates** - Simple boolean state checks (e.g., `is_empty`, `is_valid`, `is_full`)
-5. **Geometric operations** - All other methods including conversions, computations, containment checks, set operations, normalization, and distance calculations
-6. **Vector operations** - Methods from the Vector base class (e.g., `norm`, `norm2`, `normalize`, `dot_prod`, `cross_prod`, `angle`). Only applicable to classes that inherit from `util/math/vector.h`
-7. **Operators** - Operator overloads (e.g., `==`, `+`, `*`, comparison operators)
-8. **String representation** - `__repr__` (which also provides `__str__`), and string conversion methods like `to_string_in_degrees`
-9. **Module-level functions** - Standalone functions (e.g., trigonometric functions for S1Angle)
+3. **Constants** - Class-level constants (e.g., `S2CellId.kMaxLevel`, `S2CellId.kNumFaces`)
+4. **Properties** - Mutable and read-only properties (e.g., coordinate accessors like `x`, `y`, `lo`, `hi`)
+5. **Predicates** - Simple boolean state checks (e.g., `is_empty`, `is_valid`, `is_full`)
+6. **Geometric operations** - Conversions, computations, containment checks, set operations, normalization, and distance calculations
+7. **Traversal** - Methods for navigating hierarchies and iterating over related cells (e.g., `parent`, `child`, `children()`, `cells()`, neighbor methods)
+8. **Vector operations** - Methods from the Vector base class (e.g., `norm`, `norm2`, `normalize`, `dot_prod`, `cross_prod`, `angle`). Only applicable to classes that inherit from `util/math/vector.h`
+9. **Operators** - Operator overloads (e.g., `==`, `+`, `*`, comparison operators)
+10. **String representation** - `__repr__` (which also provides `__str__`), and string conversion methods like `to_string_in_degrees`
+11. **Module-level functions** - Standalone functions (e.g., trigonometric functions for S1Angle)
--- a/src/python/README.md
+++ b/src/python/README.md
@@ -149,7 +149,8 @@ Use the following sections to organize functions within the bindings files and t
 4. **Properties** - Mutable and read-only properties (e.g., coordinate accessors like `x`, `y`, `lo`, `hi`)
 5. **Predicates** - Simple boolean state checks (e.g., `is_empty`, `is_valid`, `is_full`)
 6. **Geometric operations** - Conversions, computations, containment checks, set operations, normalization, and distance calculations
-7. **Vector operations** - Methods from the Vector base class (e.g., `norm`, `norm2`, `normalize`, `dot_prod`, `cross_prod`, `angle`). Only applicable to classes that inherit from `util/math/vector.h`
-8. **Operators** - Operator overloads (e.g., `==`, `+`, `*`, comparison operators)
-9. **String representation** - `__repr__` (which also provides `__str__`), and string conversion methods like `to_string_in_degrees`
-10. **Module-level functions** - Standalone functions (e.g., trigonometric functions for S1Angle)
+7. **Traversal** - Methods for navigating a cell hierarchy (e.g., `parent`, `child`, neighbor methods)
+8. **Vector operations** - Methods from the Vector base class (e.g., `norm`, `norm2`, `normalize`, `dot_prod`, `cross_prod`, `angle`). Only applicable to classes that inherit from `util/math/vector.h`
+9. **Operators** - Operator overloads (e.g., `==`, `+`, `*`, comparison operators)
+10. **String representation** - `__repr__` (which also provides `__str__`), and string conversion methods like `to_string_in_degrees`
+11. **Module-level functions** - Standalone functions (e.g., trigonometric functions for S1Angle)
--- a/src/python/README.md
+++ b/src/python/README.md
@@ -76,10 +76,6 @@ The Python bindings follow the C++ API closely but with Pythonic conventions:
 - C++ functions that accept or return a vector object use a Python tuple (of length matching the vector dimension)
 - Array indexing operators (e.g., `point[0]`) are not currently supported
 
-**Iterators:**
-- Some classes support `__iter__` for traversal (e.g., `for c in S2CellId.cells(level=5):`) or expose methods that return iterables.
-- Iterables support forward iteration, `len()`, indexing, `in`, and `reversed()` unless otherwise noted.
-
 **Serialization:**
 - The C++ Encoder/Decoder serialization functions are not currently supported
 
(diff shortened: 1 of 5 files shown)


deustis:
@jmr, ready for review!

jmr:
> - `for c in cell:` / `reversed(cell)` walks the Hilbert curve

This is unclear to me.  Is this children? Leaf cells? I don't think I'd have an interface like this, but could be missing something.

deustis:
> > * `for c in cell:` / `reversed(cell)` walks the Hilbert curve
> 
> This is unclear to me. Is this children? Leaf cells? I don't think I'd have an interface like this, but could be missing something.

There are several iteration patterns in the C++ class. This self iterable binding provides the same functionality as next/prev:
https://github.com/google/s2geometry/blob/master/src/s2/s2cell_id.h#L348
```
  // Return the next/previous cell at the same level along the Hilbert curve.
  // Works correctly when advancing from one face to the next, but
  // does *not* wrap around from the last face to the first or vice versa.
  IFNDEF_SWIG([[nodiscard]]) S2CellId next() const;
  IFNDEF_SWIG([[nodiscard]]) S2CellId prev() const;
```

The functionality is very similar to Begin/End except it starts at the current cell rather than at the first cell in a level:
```
  // Iterator-style methods for traversing all the cells along the Hilbert
  // curve at a given level (across all 6 faces of the cube).  Note that the
  // end value is exclusive (just like standard STL iterators), and is not a
  // valid cell id.
  static S2CellId Begin(int level);
  static S2CellId End(int level);
```
(This one is bound as the static `cells` method)

I tried to make these iteration patterns more consistent and pythonic with `S2CellIdRange` and `S2CellIdForwardIter`. Very much open to feedback on whether we should expose them all and also on naming.

jmr:
How many more PRs do you have stacked up behind this one?  An easy way might be to remove the iteration, go on with the rest and come back to the iteration later.

src/python/s2cell_id_bindings.cc
```diff
@@ -0,0 +1,389 @@
...
+        return oss.str();
+      })
+      .def("__str__", [](S2CellId id) {
+        std::ostringstream oss;
+        oss << id;
+        return oss.str();
+      });
+
+  py::class_<S2CellIdRange>(m, "S2CellIdRange")
+      .def("__len__", &S2CellIdRange::size)
```
jmr: What's the max value here?  This needs to fit in `Py_ssize_t`.

src/python/s2cell_id_bindings.cc
```diff
@@ -0,0 +1,389 @@
...
+           "Construct a leaf cell containing the given point.\n\n"
+           "The point does not need to be normalized.")
+      .def(py::init([](const S2LatLng& ll) {
+               return S2CellId(ll);
+           }),
+           py::arg("latlng"),
+           "Construct a leaf cell containing the given S2LatLng.")
+
+      // Constants
+      .def_property_readonly_static("MAX_LEVEL",
```
jmr: How is this different from `m.attr("MAX_LEVEL") = S2CellId::kMaxLevel`?

src/python/s2cell_id_bindings.cc
```diff
@@ -0,0 +1,389 @@
...
+  }
+
+  bool contains(S2CellId cell) const {
+    return cell >= begin && cell < end &&
+           cell.level() == begin.level();
+  }
+};
+
+// Forward iterator for S2CellIdRange.
+struct S2CellIdForwardIter {
```
jmr: Should these be `module_local`?

src/python/s2cell_id_bindings.cc
```diff
@@ -0,0 +1,389 @@
...
+    } else {
+      cur = cur.prev();
+    }
+    return result;
+  }
+};
+
+}  // namespace
+
+void bind_s2cell_id(py::module& m) {
```
jmr, line 65: Would `__bool__` ever be useful? should it be equivalent to `is_valid()`?

src/python/s2cell_id_bindings.cc
```diff
@@ -0,0 +1,389 @@
...
+      })
+      .def("__str__", [](S2CellId id) {
+        std::ostringstream oss;
+        oss << id;
+        return oss.str();
+      });
+
+  py::class_<S2CellIdRange>(m, "S2CellIdRange")
+      .def("__len__", &S2CellIdRange::size)
+      .def("__getitem__", &S2CellIdRange::item, py::arg("index"))
```
jmr: what about slices?

jmr:
> The functionality is very similar to Begin/End except it starts at the current cell rather than at the first cell in a level:

So it does a third thing that I didn't guess.  `range(cell, end(cell.level))` will work here.  I think we'd want to see how often this comes up before having a special solution.

src/python/s2cell_id_bindings.cc
deustis: Will address in iteration PR

src/python/s2cell_id_bindings.cc
deustis: Switched to `def_readonly_static` to avoid the lambda.. Kept the constants on the class rather than `m.attr("MAX_LEVEL")` since they are specific to `S2CellId`. A module-level `MAX_LEVEL` would be ambiguous: `s2geometry.MAX_LEVEL`.

src/python/s2cell_id_bindings.cc
deustis: Will be addressed in the iteration follow-up PR. Yes, will make `S2CellIdForwardIter`  and `SCellIdReverseIter` module_local

src/python/s2cell_id_bindings.cc
deustis: Following the pattern of disallowing invalid states, the current bindings don't expose none or sentinel values. So __bool__ would always return true, and is_valid() is also not exposed because it would always be true.

Let me know if you prefer to make an exception here to expose none/sentinel values for this class.

src/python/s2cell_id_bindings.cc
deustis: Will be addressed in the iteration follow-up PR.

deustis:
@jmr, took your suggestion to split iteration into a follow-up. Should be ready for another review pass

src/python/s2cell_id_bindings.cc
jmr: hmm. don't we actually need `Sentinel()` for some uses? And what happens if a wrapped C++ function wanted to return `None()`?

jmr:
Ask your favorite LLM to do a code review.  Ask if the pybind11 code and generated Python bindings are idiomatic.

src/python/README.md
```diff
@@ -145,10 +145,12 @@ Use the following sections to organize functions within the bindings files and t
...
-9. **Module-level functions** - Standalone functions (e.g., trigonometric functions for S1Angle)
+3. **Constants** - Class-level constants in upper snake case (e.g., `S2CellId.MAX_LEVEL`, `S2CellId.NUM_FACES`)
+4. **Properties** - Mutable and read-only properties (e.g., coordinate accessors like `x`, `y`, `lo`, `hi`)
+5. **Predicates** - Simple boolean state checks (e.g., `is_empty`, `is_valid`, `is_full`)
+6. **Geometric operations** - Conversions, computations, containment checks, set operations, normalization, and distance calculations
+7. **Traversal** - Methods for navigating a cell hierarchy (e.g., `parent`, `child`, neighbor methods)
+8. **Vector operations** - Methods from the Vector base class (e.g., `norm`, `norm2`, `normalize`, `dot_prod`, `cross_prod`, `angle`). Only applicable to classes that inherit from `util/math/vector.h`
+9. **Operators** - Operator overloads (e.g., `==`, `+`, `*`, comparison operators)
+10. **String representation** - `__repr__` (which also provides `__str__`), and string conversion methods like `to_string_in_degrees`
+11. **Module-level functions** - Standalone functions (e.g., trigonometric functions for S1Angle)
```
jmr: This is getting too long and too hard to see the diffs. Use "1." everywhere. 

https://google.github.io/styleguide/docguide/style.html#use-lazy-numbering-for-long-lists

I will update the existing items.

deustis:
Idiomaticity review (from Claude, per jmr's suggestion):

**Overall: the bindings are well-structured.** Naming conventions (snake_case methods, UPPER_SNAKE constants), exception types (ValueError throughout), constructor overloading, and the tuple-vs-list distinction for fixed/variable-size neighbor returns are all idiomatic. Docstrings are appropriate and consistent with the other binding files.

Three items worth considering:

**1. `face`, `level`, `pos` could be properties instead of methods.**
The project convention uses properties for cheap O(1) accessors: `S2Point.x`, `S1Angle.radians`, `S2LatLng.lat`, and `cell.id` is already a property in this file. Having `cell.id` as a property but `cell.level()` as a method is inconsistent. `cell.face`, `cell.level`, `cell.pos` would match both the project convention and Python user expectations.

**2. `__repr__` format is not eval-able.**
`repr(cell)` returns `S2CellId(0/)` which looks like it should round-trip through `eval()` but can't. Python convention says repr should either be eval-able or use angle-bracket notation. Using the numeric id (`S2CellId(1152921504606846976)`) would actually round-trip via the uint64 constructor. However, the current approach is consistent with the other bindings in this project (S2Point, S1Angle, S2LatLng all wrap C++ `operator<<` output), so this may not be worth changing unless all bindings are updated together.

**3. `is_leaf`/`is_face` — methods vs properties.**
These are cheap bit-checks on an immutable value, so they'd work as properties. But `S1Angle.is_normalized()` and similar predicates in the project are methods. Keeping them as methods is consistent and defensible.

src/python/s2cell_id_bindings.cc
deustis: I looked at all the uses of `None()` and `Sentinel()` across the codebase. They fall into three patterns, each with a natural Pythonic translation that avoids exposing the raw sentinels:

**Pattern 1: Error returns from parsing** — `FromToken()` and `FromDebugString()` return `None()` on malformed input. The binding already handles this: `from_token()` raises `ValueError` instead.

**Pattern 2: "No value" sentinel in query/index results** — `S2ClosestCellQuery::Result` defaults to `None()` and checks `is_empty()` via `cell_id_ == None()`. `S2CellIndex` and `S2PointIndex` use `None()` as field defaults. When binding these classes, the Python wrapper would return `Optional[S2CellId]` (Python `None`) instead.

**Pattern 3: Past-the-end iterator marker** — `S2CellIterator`, `S2CellRangeIterator`, `S2ShapeIndex` iterators return `Sentinel()` when done. Python bindings would use the `StopIteration` protocol instead.

Given that each use case maps to a standard Python idiom, I'd propose keeping `None()`/`Sentinel()` out of the Python API and letting each downstream binding translate. Happy to reconsider if there's a use case I'm missing.

deustis:
> 1. face, level, pos could be properties instead of methods. The project convention uses properties for cheap O(1) accessors: S2Point.x, S1Angle.radians, S2LatLng.lat, and cell.id is already a property in this file. Having cell.id as a property but cell.level() as a method is inconsistent. cell.face, cell.level, cell.pos would match both the project convention and Python user expectations.

Switched to properties for those methods, skipped the other suggestions.

deustis:
@jmr, I think this is ready for another round

src/python/s2cell_id_bindings.cc
```diff
@@ -0,0 +1,275 @@
...
+           "Return the edge length of this cell in (i,j)-space")
+      .def_static("get_size_ij_for_level", [](int level) {
+               MaybeThrowLevelOutOfRange(level, 0, S2CellId::kMaxLevel);
+               return S2CellId::GetSizeIJ(level);
+           }, py::arg("level"),
+           "Return the edge length in (i,j)-space of cells at the given level")
+      .def("get_size_st",
+           py::overload_cast<>(&S2CellId::GetSizeST, py::const_),
+           "Return the edge length of this cell in (s,t)-space")
+      .def_static("get_size_st_for_level", [](int level) {
```
jmr, line 159: needs test

src/python/s2cell_id_test.py
```diff
@@ -0,0 +1,356 @@
...
+        self.assertEqual(len(neighbors), 4)
+
+    def test_get_vertex_neighbors(self):
+        cell = s2.S2CellId.from_face(0).child(0).child(1).child(2)
+        neighbors = cell.get_vertex_neighbors(1)
+        self.assertGreaterEqual(len(neighbors), 3)
+        self.assertLessEqual(len(neighbors), 4)
+
+    def test_get_vertex_neighbors_level_too_high_raises(self):
+        cell = s2.S2CellId.from_face(0)
```
jmr: have separate test cases for face and level, since you have separate checks for these

src/python/README.md
```diff
@@ -145,9 +145,11 @@ Use the following sections to organize functions within the bindings files and t
 
 1. **Constructors** - Default constructors and constructors with parameters
 1. **Factory methods** - Static factory methods (e.g., `from_degrees`, `from_radians`, `zero`, `invalid`)
+1. **Constants** - Class-level constants in upper snake case (e.g., `S2CellId.MAX_LEVEL`, `S2CellId.NUM_FACES`)
```
jmr, line 147: constants first?

src/python/s2cell_id_bindings.cc
deustis: Done. Added `test_get_size_st_for_level` covering level 0 (full unit interval) and level 30.

src/python/s2cell_id_test.py
deustis: Done.  Split into `test_get_vertex_neighbors_face_cell_raises` (exercises `MaybeThrowIfFace`) and `test_get_vertex_neighbors_level_out_of_range_raises` (exercises `MaybeThrowLevelOutOfRange` with a level-3 cell).

src/python/README.md
deustis: Done. Moved Constants before Factory methods in the README, matching the order already used in the binding file.

jmr (approved):

deustis:
@jmr, friendly ping on this PR?

PR #607, opened 2026-05-10
workflows: add lint presubmits for BUILD files

### LINKED ISSUE #549: lint presubmits
If there's an easy way to set up linters for the C++/Python/BUILD formatting, we should do that.

abhinavgautam01:
Fixes #549

## Summary
Adds a dedicated **lint** GitHub Actions workflow so C++, Python, and Bazel **BUILD** files get consistent formatting checks before merge. This directly addresses the request in #549 for lightweight presubmit tooling.
## What runs in CI
- **BUILD / Starlark** — **`buildifier`** (pinned **v8.5.1**) with **`-mode=check`** over `BUILD` / `BUILD.bazel`, `*.bzl`, `MODULE.bazel`, and `WORKSPACE` files under the
 repo.
- **Python** — **`ruff format --check`** (pinned **ruff 0.11.13**) over all **tracked** `*.py` files (`git ls-files '*.py'`). In-repo Python sources were formatted so this check passes on a clean tree.
- **C++ / headers** — **`clang-format-18`** with **`--dry-run --Werror`**, applied only to **C/C++ source files touched** in each PR/push range (identified via `git diff`), not to the entire `src/` tree.
## Why C++ is diff-scoped only
A full-repository `clang-format` check currently reports many violations on upstream. Turning that into a mandatory gate would block unrelated contributions until a large, dedicated formatting-only change lands. Scoped checks match the project’s documented workflow (formatting touched code with tooling such as **`git clang-format`**) while
still enforcing style on **new and modified** `.cc`/`.h` files.
## Ancillary fixes
- **`src/python/BUILD.bazel`** — **`buildifier`** formatting and cleanup (including correcting a malformed dependency line).
- **`pyproject.toml`** — minimal **`[tool.ruff]`** section with **`target-version = "py310"`** alongside existing `requires-python`.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,40 @@
+name: lint
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    timeout-minutes: 1
+
+    steps:
+      # Default clone depth is enough here: we only run buildifier on the
+      # checked-out tree (no PR-range git diffs in this workflow).
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: Install buildifier
+        run: |
+          set -u
+          BUILDIFIER_VERSION=8.5.1
+          BUILDIFIER_SHA256=887377fc64d23a850f4d18a077b5db05b19913f4b99b270d193f3c7334b5a9a7
+          BUILDIFIER_DL=/tmp/buildifier
+          curl -fsSL -o "${BUILDIFIER_DL}" \
+            "https://github.com/bazelbuild/buildtools/releases/download/v${BUILDIFIER_VERSION}/buildifier-linux-amd64"
+          echo "${BUILDIFIER_SHA256}  ${BUILDIFIER_DL}" | sha256sum --check -
+          chmod +x "${BUILDIFIER_DL}"
+          sudo mv "${BUILDIFIER_DL}" /usr/local/bin/buildifier
+
+      - name: Check Starlark / BUILD formatting (buildifier)
+        run: buildifier -r -mode=check .


.github/workflows/lint.yml
```diff
@@ -0,0 +1,96 @@
...
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+  workflow_dispatch:
+
+concurrency:
+  group: lint-${{ github.workflow }}-${{ github.head_ref || github.ref }}
```
jmr: `lint-` is redundant

.github/workflows/lint.yml
```diff
@@ -0,0 +1,96 @@
...
+        with:
+          fetch-depth: 0
+
+      - name: Install clang-format and buildifier
+        run: |
+          set -euxo pipefail
+          sudo apt-get update -qq
+          sudo apt-get install -y -qq clang-format-18
+          curl -fsSL -o /tmp/buildifier https://github.com/bazelbuild/buildtools/releases/download/v8.5.1/buildifier-linux-amd64
+          chmod +x /tmp/buildifier
```
jmr: `echo "<expected_sha256>  /tmp/buildifier" | sha256sum --check`

.github/workflows/lint.yml
```diff
@@ -0,0 +1,96 @@
...
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          fetch-depth: 0
```
jmr: Not sure about this.  Add a comment for why.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,96 @@
...
+        with:
+          python-version: "3.12"
+
+      - name: Install ruff
+        run: pip install ruff==0.11.13
+
+      - name: Check Starlark / BUILD formatting (buildifier)
+        run: |
+          set -euxo pipefail
+          find . -type f \
```
jmr: `buildifier -r -mode=check .`

.github/workflows/lint.yml
```diff
@@ -0,0 +1,96 @@
...
+  lint:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          fetch-depth: 0
+
+      - name: Install clang-format and buildifier
```
jmr: First add just `buildifier`.  It's closest to being mergeable.  The others will need more work.

.github/workflows/lint.yml
abhinavgautam01: Done — dropped the redundant lint- prefix; the concurrency group now matches the other workflows...

.github/workflows/lint.yml
abhinavgautam01: Added a pinned SHA256 for v8.5.1 linux-amd64 and echo "…  /tmp/buildifier" | sha256sum --check - after the download...

.github/workflows/lint.yml
abhinavgautam01: Removed it and left a short comment: we only need the default checkout for buildifier (no diff-based steps in this workflow anymore)...

.github/workflows/lint.yml
abhinavgautam01: Switched the check to buildifier -r -mode=check . as requested...

.github/workflows/lint.yml
abhinavgautam01: scoped this workflow to buildifier only for now; dropped the clang-format and ruff steps so the rest can iterate separately...

.github/workflows/lint.yml
```diff
@@ -0,0 +1,42 @@
...
+    timeout-minutes: 15
+
+    steps:
+      # Default clone depth is enough here: we only run buildifier on the
+      # checked-out tree (no PR-range git diffs in this workflow).
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: Install buildifier
+        run: |
+          set -euxo pipefail
```
jmr: `set -u` is enough

abhinavgautam01: updated both run blocks to use set -u only. GitHub Actions’ default bash already runs with -e and pipefail, so set -euxo pipefail was redundant; set -u still catches unset variables during the install and buildifier steps...

.github/workflows/lint.yml
```diff
@@ -0,0 +1,42 @@
...
+          BUILDIFIER_DL=/tmp/buildifier
+          curl -fsSL -o "${BUILDIFIER_DL}" \
+            "https://github.com/bazelbuild/buildtools/releases/download/v${BUILDIFIER_VERSION}/buildifier-linux-amd64"
+          echo "${BUILDIFIER_SHA256}  ${BUILDIFIER_DL}" | sha256sum --check -
+          chmod +x "${BUILDIFIER_DL}"
+          sudo mv "${BUILDIFIER_DL}" /usr/local/bin/buildifier
+
+      - name: Check Starlark / BUILD formatting (buildifier)
+        run: |
+          set -euxo pipefail
```
jmr: not needed

jmr (changes requested):
Revert all the Python changes, they're no longer relevant.

.github/workflows/lint.yml
abhinavgautam01: dropped the second multiline run / set -u block and run buildifier -r -mode=check . as a one-line run — no variables there, so the shell wrapper wasn’t doing anything useful...

abhinavgautam01:
> Revert all the Python changes, they're no longer relevant.

reverted all Python-side edits on this branch (pyproject.toml, setup.py, doc/examples/term_index.py, src/python/BUILD.bazel, and the touched *_test.py files) back to master, since the workflow is buildifier-only now and those changes aren’t relevant to the PR...

.github/workflows/lint.yml
```diff
@@ -0,0 +1,40 @@
...
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
```
jmr: This took 4 seconds to run. How about a one minute timeout?

abhinavgautam01: updated...!

jmr:
Merge or rebase onto HEAD. The BUILD.bazel changes should already be there.

jmr (approved):

jmr:
You can do the C++ changes next.  Python requires more thought.  ruff seems better, but we'd need to make the formats match.

abhinavgautam01:
> You can do the C++ changes next. Python requires more thought. ruff seems better, but we'd need to make the formats match.

Will do C++ next; I’ll hold off on Python/ruff until we’ve pinned down format parity. Thanks for the direction...

Issue #395, opened 2024-11-30
s2loop_measures_test GetSignedArea.ErrorAccumulation fails on arm64 macOS


jmr:
Unclear if this is a problem with the allowed error threshold or the code.

```
[ RUN      ] GetSignedArea.ErrorAccumulation
.../s2geometry/src/s2/s2loop_measures_test.cc:286: Failure
The difference between expected_area and actual_area is 3.5813685350660762e-11, which exceeds allowed_error, where
expected_area evaluates to 0.78539816339744795,
actual_area evaluates to 0.78539816336163426, and
allowed_error evaluates to 3.4879033766936103e-11.

[  FAILED  ] GetSignedArea.ErrorAccumulation (343 ms)
```

```
% c++ --version
Apple clang version 15.0.0 (clang-1500.1.0.2.5)
Target: arm64-apple-darwin22.6.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

PR #608, opened 2026-05-10
workflows: Retry flaky tests 3 times in ctest

### LINKED ISSUE #521: Add flaky test support to CMakeLists.txt
I just had to rerun the workflows 3 times to change README.md.

We can do something like add a "flaky" property with [`set_tests_properties`](https://cmake.org/cmake/help/latest/command/set_tests_properties.html), then run the non-flaky tests once and the flaky ones with 3 retries.

abhinavgautam01:
Fixes #521
## Summary
Adds CMake/CTest **`LABELS flaky`** support for known-flaky **`s2loop_measures_test`** discoveries and splits GitHub Actions **`ctest`** into a strict phase plus a flaky phase with retries.
## CMake
- For **`s2loop_measures_test`** only, pass **`PROPERTIES LABELS "flaky"`** to **`gtest_discover_tests`**, so every discovered case in that binary is labelled (see README “Status” / **#395**, especially **`GetSignedArea.ErrorAccumulation`**-style numeric sensitivity).
## CI
- **Normal + big-endian CMake jobs:**
  `ctest … -LE flaky` then
  `ctest … -L flaky --repeat until-pass:3`
  (big-endian keeps the existing **`-E "$EXCLUDE"`** filter on both invocations.)
## Local / maintenance
Roughly mirrors CI:
```bash
ctest --output-on-failure -LE flaky
ctest --output-on-failure -L flaky --repeat until-pass:3
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -65,12 +65,8 @@ jobs:
         run: |
           set -u
           # Join one-regex-per-line list (same alternation as a single POSIX ERE).
-          # macOS paste(1) has no GNU long options; Linux uses --serial/--delimiters.
-          if [ '${{ runner.os }}' = 'macOS' ]; then
-            FLAKY="$(paste -s -d '|' ../flaky_tests.txt)"
-          else
-            FLAKY="$(paste --serial --delimiters='|' ../flaky_tests.txt)"
-          fi
+          # macOS doesn't support long options.
+          FLAKY="$(paste -s -d '|' ../flaky_tests.txt)"
           ctest --parallel ${{ env.NUM_CORES }} --output-on-failure \
             --exclude-regex "${FLAKY}"
           ctest --parallel ${{ env.NUM_CORES }} --output-on-failure \
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -62,7 +62,10 @@ jobs:
       - run: cmake --build . --parallel ${{ env.NUM_CORES }}
         working-directory: build/
       - if: always()
-        run: ctest --parallel ${{ env.NUM_CORES }} --output-on-failure
+        run: |
+          set -euxo pipefail
+          ctest --parallel "${NUM_CORES}" --output-on-failure -LE flaky
+          ctest --parallel "${NUM_CORES}" --output-on-failure -L flaky --repeat until-pass:3
         working-directory: build/
   cmake-big-endian:
     runs-on: ubuntu-24.04-arm
@@ -116,6 +119,7 @@ jobs:
       - if: always()
         name: Run tests
         run: |
+          set -euxo pipefail
           # Known failures due to unimplemented big-endian encoding/decoding.
           EXCLUDE="EncodedS2PointVectorTest\.(Empty|OnePoint|AllExceptionsBlock|RoundtripEncodingFast)"
           EXCLUDE+="|EncodedS2ShapeIndex\.(OneEdge|RegularLoops|OverlappingPointClouds|OverlappingPolylines|OverlappingLoops|JavaByteCompatibility)"
@@ -127,7 +131,8 @@ jobs:
           EXCLUDE+="|FastEncodeTaggedShapes\.MixedShapes"
           EXCLUDE+="|DecodeTaggedShapes\.(DecodeFromByteString|DecodeFromEncoded)"
           EXCLUDE+="|OutgoingEdgeButNoIncomingEdge"
-          ctest --parallel ${{ env.NUM_CORES }} --output-on-failure -E "$EXCLUDE"
+          ctest --parallel "${NUM_CORES}" --output-on-failure -LE flaky -E "$EXCLUDE"
+          ctest --parallel "${NUM_CORES}" --output-on-failure -L flaky -E "$EXCLUDE" --repeat until-pass:3
         working-directory: build/
   bazel:
     runs-on: ubuntu-latest
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -64,7 +64,8 @@ jobs:
       - if: always()
         run: |
           set -euxo pipefail
-          FLAKY="$(tr -d '\n\r' < s2_ctest_flaky_tests.regex)"
+          # Join one-regex-per-line list (same alternation as a single POSIX ERE).
+          FLAKY="$(paste -s -d '|' ../flaky_tests.txt)"
           ctest --parallel ${{ env.NUM_CORES }} --output-on-failure \
             --exclude-regex "${FLAKY}"
           ctest --parallel ${{ env.NUM_CORES }} --output-on-failure \
@@ -135,7 +136,7 @@ jobs:
           EXCLUDE+="|FastEncodeTaggedShapes\.MixedShapes"
           EXCLUDE+="|DecodeTaggedShapes\.(DecodeFromByteString|DecodeFromEncoded)"
           EXCLUDE+="|OutgoingEdgeButNoIncomingEdge"
-          FLAKY="$(tr -d '\n\r' < s2_ctest_flaky_tests.regex)"
+          FLAKY="$(paste -s -d '|' ../flaky_tests.txt)"
           EXCLUDE_MAIN="${EXCLUDE}|${FLAKY}"
           ctest --parallel ${{ env.NUM_CORES }} --output-on-failure \
             --exclude-regex "${EXCLUDE_MAIN}"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -63,9 +63,14 @@ jobs:
         working-directory: build/
       - if: always()
         run: |
-          set -euxo pipefail
+          set -u
           # Join one-regex-per-line list (same alternation as a single POSIX ERE).
-          FLAKY="$(paste -s -d '|' ../flaky_tests.txt)"
+          # macOS paste(1) has no GNU long options; Linux uses --serial/--delimiters.
+          if [ '${{ runner.os }}' = 'macOS' ]; then
+            FLAKY="$(paste -s -d '|' ../flaky_tests.txt)"
+          else
+            FLAKY="$(paste --serial --delimiters='|' ../flaky_tests.txt)"
+          fi
           ctest --parallel ${{ env.NUM_CORES }} --output-on-failure \
             --exclude-regex "${FLAKY}"
           ctest --parallel ${{ env.NUM_CORES }} --output-on-failure \
@@ -124,7 +129,7 @@ jobs:
       - if: always()
         name: Run tests
         run: |
-          set -euxo pipefail
+          set -u
           # Known failures due to unimplemented big-endian encoding/decoding.
           EXCLUDE="EncodedS2PointVectorTest\.(Empty|OnePoint|AllExceptionsBlock|RoundtripEncodingFast)"
           EXCLUDE+="|EncodedS2ShapeIndex\.(OneEdge|RegularLoops|OverlappingPointClouds|OverlappingPolylines|OverlappingLoops|JavaByteCompatibility)"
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -521,7 +521,16 @@ if (BUILD_TESTS)
     # Compared to `add_test()`, displays progress and success/failure of each
     # test case rather than just the binary, so we see
     # `S2Cell.TestSubdivide ... PASSED`.  Also lists disabled tests.
-    gtest_discover_tests(${test})
+    #
+    # Mark s2loop_measures_test as flaky: some cases (notably
+    # GetSignedArea.ErrorAccumulation) are numerically sensitive on some
+    # platforms; see README "Status" and #395.  CI can split runs with
+    # `ctest -LE flaky` and `ctest -L flaky --repeat until-pass:3` (#521).
+    if (test STREQUAL "s2loop_measures_test")
```
jmr: Do something like this:

```
  set(FLAKY_TESTS s2loop_measures_test)
  # ...
  if (test IN_LIST FLAKY_TESTS)
    gtest_discover_tests(${test} PROPERTIES LABELS "flaky")
  else()
    gtest_discover_tests(${test})
  endif()
```

There are more flaky tests.  Look for anything with `ConfidenceBound` and look at the issues.  All test failures there are flaky since I don't think we have anything that always fails.

.github/workflows/build.yml
```diff
@@ -127,7 +131,8 @@ jobs:
           EXCLUDE+="|FastEncodeTaggedShapes\.MixedShapes"
           EXCLUDE+="|DecodeTaggedShapes\.(DecodeFromByteString|DecodeFromEncoded)"
           EXCLUDE+="|OutgoingEdgeButNoIncomingEdge"
-          ctest --parallel ${{ env.NUM_CORES }} --output-on-failure -E "$EXCLUDE"
+          ctest --parallel "${NUM_CORES}" --output-on-failure -LE flaky -E "$EXCLUDE"
```
jmr: `${{ env.NUM_CORES }}` got changed to `"${NUM_CORES}"`. Change it back.

.github/workflows/build.yml
```diff
@@ -127,7 +131,8 @@ jobs:
...
+          ctest --parallel "${NUM_CORES}" --output-on-failure -L flaky -E "$EXCLUDE" --repeat until-pass:3
```
jmr: `-E` was sort of clear when it was the only option.  Please use the long forms `--exclude-regex` etc now that there are three different options.

CMakeLists.txt
abhinavgautam01: Done. I added a set(FLAKY_TESTS …) block before the foreach and now use if (test IN_LIST FLAKY_TESTS) around gtest_discover_tests, as you suggested...

.github/workflows/build.yml
abhinavgautam01: Reverted those ctest lines to use ${{ env.NUM_CORES }} again instead of the shell "${NUM_CORES}" so they stay consistent with the Set NUM_CORES step.

.github/workflows/build.yml
abhinavgautam01: Updated the flaky / exclude invocations to use --label-exclude, --label-regex, and --exclude-regex instead of -LE, -L, and -E, while keeping --parallel, --output-on-failure and --repeat until-pass:3 as before.

CMakeLists.txt
```diff
@@ -497,6 +497,17 @@ if (BUILD_TESTS)
 
   enable_testing()
 
+  # Tests that use randomized / statistical bounds (e.g. ConfidenceBound) or
+  # that have open flaky reports — mark so CI can ctest --label-exclude/--repeat
+  # (#521). Not expected to fail every run on a healthy tree.
+  set(FLAKY_TESTS
+      s2loop_measures_test
```
jmr: sort the file list

CMakeLists.txt
```diff
@@ -497,6 +497,17 @@ if (BUILD_TESTS)
 
   enable_testing()
 
+  # Tests that use randomized / statistical bounds (e.g. ConfidenceBound) or
+  # that have open flaky reports — mark so CI can ctest --label-exclude/--repeat
+  # (#521). Not expected to fail every run on a healthy tree.
+  set(FLAKY_TESTS
```
jmr: Are these same tests marked flaky in BUILD.bazel?

CMakeLists.txt
abhinavgautam01: done, FLAKY_TESTS is sorted alphabetically now and the comment says to keep it that way...!

CMakeLists.txt
abhinavgautam01: They weren’t before; I added flaky = True on the same six cc_test targets in src/BUILD.bazel so CMake’s flaky label and Bazel line up. s2polygon_test still has shard_count = 5.

CMakeLists.txt
```diff
@@ -497,6 +497,17 @@ if (BUILD_TESTS)
...
+      s2loop_measures_test
+      s2closest_edge_query_test
+      s2furthest_edge_query_test
+      s2convex_hull_query_test
+      s2polygon_test
+      s2predicates_test)
```
jmr: Wait, this is just files, but `EXCLUDE` for the big-endian CI has the actual test cases.  Use the test cases that are flaky, not the entire file.

CMakeLists.txt
```diff
@@ -497,6 +497,17 @@ if (BUILD_TESTS)
 
   enable_testing()
 
+  # Tests that use randomized / statistical bounds (e.g. ConfidenceBound) or
+  # that have open flaky reports — mark so CI can ctest --label-exclude/--repeat
+  # (#521). Not expected to fail every run on a healthy tree.
```
jmr: No need to mention #521.

CMakeLists.txt
abhinavgautam01: agreed, dropped whole-binary LABELS / FLAKY_TESTS. There’s now a POSIX regex of specific flaky googletests (same spirit as big-endian EXCLUDE) written as build/s2_ctest_flaky_tests.regex, and CI uses ctest --exclude-regex / --tests-regex with --repeat until-pass on those cases only.

CMakeLists.txt
abhinavgautam01: removed the #521 reference from the CMake comment above that block...

CMakeLists.txt
```diff
@@ -497,6 +497,17 @@ if (BUILD_TESTS)
 
   enable_testing()
 
+  # Names of flaky gtest cases (CTest matches substrings against this POSIX
+  # extended regex — same style as cmake-big-endian's EXCLUDE). CI passes it to
+  # ctest --exclude-regex / --tests-regex with --repeat until-pass.
+  set(_s2_ctest_flaky_regex
```
jmr: What's the advantage of writing a file in CMakeLists.txt?  Check it in as `flaky_tests.txt`, one regex per line, then use `FLAKY="$(paste --serial --delimiters='|' flaky_tests.txt)"` in build.yml.

jmr (changes requested):

CMakeLists.txt
abhinavgautam01: Checked in flaky_tests.txt at the repo root (one POSIX ERE per line) and build the ctest pattern with paste -s -d '|' ../flaky_tests.txt from build/ (same alternation as the old single string). Removed the CMake file(WRITE …) logic so the list stays version-controlled and reviewers can edit it without touching CMakeLists.txt. Used -s -d '|' instead of --serial --delimiters so it works on macOS runners too (BSD paste doesn’t support the long GNU options).

flaky_tests.txt
```diff
@@ -0,0 +1,6 @@
+GetSignedArea\.ErrorAccumulation
```
jmr: sort

.github/workflows/build.yml
```diff
@@ -62,7 +62,15 @@ jobs:
       - run: cmake --build . --parallel ${{ env.NUM_CORES }}
         working-directory: build/
       - if: always()
-        run: ctest --parallel ${{ env.NUM_CORES }} --output-on-failure
+        run: |
+          set -euxo pipefail
+          # Join one-regex-per-line list (same alternation as a single POSIX ERE).
+          FLAKY="$(paste -s -d '|' ../flaky_tests.txt)"
```
jmr, line 69: Use long options for clarity.

.github/workflows/build.yml
```diff
@@ -62,7 +62,15 @@ jobs:
       - run: cmake --build . --parallel ${{ env.NUM_CORES }}
         working-directory: build/
       - if: always()
-        run: ctest --parallel ${{ env.NUM_CORES }} --output-on-failure
+        run: |
+          set -euxo pipefail
```
jmr: set -u is enough

jmr (changes requested):

flaky_tests.txt
abhinavgautam01: Sorted flaky_tests.txt lexicographically.

.github/workflows/build.yml
abhinavgautam01: Switched to paste --serial --delimiters='|' on Linux (including big-endian). For the main cmake matrix, macOS still uses paste -s -d '|' because BSD paste doesn’t support those long options.

.github/workflows/build.yml
abhinavgautam01: Replaced set -euxo pipefail with set -u on these steps; run already gets -e / pipefail from the default shell.

.github/workflows/build.yml
```diff
@@ -62,7 +62,20 @@ jobs:
       - run: cmake --build . --parallel ${{ env.NUM_CORES }}
         working-directory: build/
       - if: always()
-        run: ctest --parallel ${{ env.NUM_CORES }} --output-on-failure
+        run: |
+          set -u
+          # Join one-regex-per-line list (same alternation as a single POSIX ERE).
+          # macOS paste(1) has no GNU long options; Linux uses --serial/--delimiters.
```
jmr: I didn't realize macOS didn't have the long options.  Then use the short ones with a comment
`# macOS doesn't support long options.`

.github/workflows/build.yml
```diff
@@ -127,7 +141,14 @@ jobs:
           EXCLUDE+="|FastEncodeTaggedShapes\.MixedShapes"
           EXCLUDE+="|DecodeTaggedShapes\.(DecodeFromByteString|DecodeFromEncoded)"
           EXCLUDE+="|OutgoingEdgeButNoIncomingEdge"
-          ctest --parallel ${{ env.NUM_CORES }} --output-on-failure -E "$EXCLUDE"
+          FLAKY="$(paste --serial --delimiters='|' ../flaky_tests.txt)"
```
jmr, line 140: This one can stay like this.

.github/workflows/build.yml
abhinavgautam01: Dropped the Linux/macOS if branch in the cmake job and always paste -s -d '|' there, with # macOS doesn't support long options. above it.

.github/workflows/build.yml
abhinavgautam01: Left cmake-big-endian as FLAKY="$(paste --serial --delimiters='|' ../flaky_tests.txt)" — unchanged on Linux.

.github/workflows/build.yml
```diff
@@ -127,7 +137,14 @@ jobs:
...
+          EXCLUDE_MAIN="${EXCLUDE}|${FLAKY}"
```
jmr: A variable that's used once isn't making anything clearer.  I was just wondering what `EXCLUDE_MAIN` was and where it came from.

abhinavgautam01: Dropped EXCLUDE_MAIN — the first pass now uses --exclude-regex "${EXCLUDE}|${FLAKY}" directly so it’s clear we’re excluding both the big-endian EXCLUDE list and the flaky regex list built from flaky_tests.txt.

jmr (approved):

jmr:
Thanks! People were always confused about why random tests failed.

abhinavgautam01:
> Thanks! People were always confused about why random tests failed.

thank you and thanks for walking through the review...the extra guidance made it much clearer how to surface the flaky cases without papering over real failures...!

PR #610, opened 2026-05-10
build: sync Bazel CXX warning flags with CMake

### LINKED ISSUE #458: warning options inconsistent
Warning options should be the same between bazel, cmake, and google3.

abhinavgautam01:
Fixes #458
Add the same **non-MSVC** C++ compile options already applied in **CMake** to **Bazel** via `src/.bazelrc` (`-Wno-attributes`, `-Wno-deprecated-declarations`, `-Wno-nullability-completeness`, `-Wno-psabi`, `-fsized-deallocation`), with a short note in **CMakeLists** to keep the two in sync. MSVC / `cl` Bazel builds are not in CI here; maintainers on Windows can override via a local/user config if needed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -119,23 +119,30 @@ if (MSVC)
     # Make sure Windows doesn't define min/max macros that interfere with STL
     add_definitions(-DNOMINMAX)
 else()
-    # Avoid megabytes of warnings like:
-    # util/math/vector.h:178:16: warning: optimization attribute on
-    # ‘double sqrt(double)’ follows definition but the attribute doesn’t
-    # match [-Wattributes]
-    add_definitions(-Wno-attributes)
-    add_definitions(-Wno-deprecated-declarations)
-    add_definitions(-Wno-nullability-completeness)
-    # Suppress noisy AArch64 ABI notes about parameter passing changes in
-    # GCC 10.1.  Irrelevant since we don't mix objects from old compilers.
-    # TODO: Use check_cxx_compiler_flag(-Wno-psabi) to guard this.
-    add_compile_options(-Wno-psabi)
-    # Some files use sized deallocation, which should be enabled by
-    # default for C++14 and later.  There appears to be a bug with clang
-    # < 19 causing this not to be enabled.
-    # https://github.com/google/s2geometry/issues/411#issuecomment-2726949607
-    # This can be removed when clang19 is the minimum supported version.
-    add_compile_options(-fsized-deallocation)
+  # OSS #458: silence only observed warning churn, guarded by toolchain support.
+
+  check_cxx_compiler_flag(-Wno-attributes S2_CXX_SUPPORTS_WNO_ATTRIBUTES)
+  if (S2_CXX_SUPPORTS_WNO_ATTRIBUTES)
+    add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-Wno-attributes>)
+  endif()
+
+  check_cxx_compiler_flag(-Wno-nullability-completeness
+                          S2_CXX_SUPPORTS_WNO_NULLABILITY_COMPLETENESS)
+  if (S2_CXX_SUPPORTS_WNO_NULLABILITY_COMPLETENESS)
+    add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-Wno-nullability-completeness>)
+  endif()
+
+  check_cxx_compiler_flag(-Wno-psabi S2_CXX_SUPPORTS_WNO_PSABI)
+  if (S2_CXX_SUPPORTS_WNO_PSABI)
+    add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-Wno-psabi>)
+  endif()
+
+  if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+    check_cxx_compiler_flag(-fsized-deallocation S2_CXX_SUPPORTS_FSIZED_DEALLOCATION)
+    if (S2_CXX_SUPPORTS_FSIZED_DEALLOCATION)
+      add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-fsized-deallocation>)
+    endif()
+  endif()
 endif()
 
 # If OpenSSL is installed in a non-standard location, configure with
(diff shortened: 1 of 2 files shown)


src/.bazelrc
```diff
@@ -3,6 +3,14 @@ common --enable_bzlmod
 common --cxxopt=-std=c++20
 common --announce_rc
 
+# CXX warning flags aligned with CMakeLists.txt non-MSVC toolchain (OSS #458).
```
jmr: In `CMakeLists.txt`, we should really be using `check_cxx_compiler_flag` and here we should have `--config=gcc`/`--config=clang` with the relevant flags for that compiler.

Let's only add suppressions we actually observe and want to suppress.

abhinavgautam01: thanks done...
CMake now gates each flag with `check_cxx_compiler_flag`, and `-fsized-deallocation` only goes on Clang. Bazel uses `--config=gcc` / `--config=clang` (**ci pulls in gcc), and I dropped the old global `build --cxxopt` lines and `Wno-deprecated`** so we only keep the suppressions we’d actually called out...!

src/.bazelrc
```diff
@@ -3,10 +3,27 @@ common --enable_bzlmod
...
 
+# OSS #458: compiler-specific warning/c++ flags (keep in conceptual sync with the
+# `check_cxx_compiler_flag` gate in CMakeLists.txt). Typical usage:
+#   bazel test --config=dev --config=clang   # AppleClang / explicit Clang
+#   bazel test --config=dev --config=gcc     # host GCC on Linux CI
+#
+# Keep these minimal: entries should correspond to churn we deliberately silence.
+
+# Host GCC toolchains (typical ubuntu-latest CI).
+build:gcc --cxxopt=-Wno-attributes
```
jmr, line 11: The important ones I see are `-Wno-deprecated-declarations` and `-Wno-sign-compare`.

https://github.com/google/s2geometry/actions/runs/25542128110/job/74970296091

CMakeLists.txt
```diff
@@ -119,23 +119,30 @@ if (MSVC)
...
     add_definitions(-DNOMINMAX)
 else()
-    # Avoid megabytes of warnings like:
-    # util/math/vector.h:178:16: warning: optimization attribute on
-    # ‘double sqrt(double)’ follows definition but the attribute doesn’t
-    # match [-Wattributes]
-    add_definitions(-Wno-attributes)
-    add_definitions(-Wno-deprecated-declarations)
-    add_definitions(-Wno-nullability-completeness)
-    # Suppress noisy AArch64 ABI notes about parameter passing changes in
```
jmr, line 129: Please keep the comments.

CMakeLists.txt
```diff
@@ -119,23 +119,30 @@ if (MSVC)
...
-    # GCC 10.1.  Irrelevant since we don't mix objects from old compilers.
-    # TODO: Use check_cxx_compiler_flag(-Wno-psabi) to guard this.
-    add_compile_options(-Wno-psabi)
-    # Some files use sized deallocation, which should be enabled by
-    # default for C++14 and later.  There appears to be a bug with clang
-    # < 19 causing this not to be enabled.
-    # https://github.com/google/s2geometry/issues/411#issuecomment-2726949607
-    # This can be removed when clang19 is the minimum supported version.
-    add_compile_options(-fsized-deallocation)
+  # OSS #458: silence only observed warning churn, guarded by toolchain support.
```
jmr: You don't need to mention this and "OSS #458" is not a style we use.

src/.bazelrc
```diff
@@ -3,10 +3,27 @@ common --enable_bzlmod
...
+build:clang --cxxopt=-Wno-nullability-completeness
+build:clang --cxxopt=-fsized-deallocation
+
 # Faster tests than fastbuild (-O0).
 build:dev --copt=-O1
 
-# CI-specific flags
+# CI-specific flags (also select the GCC-aligned warning config).
 common:ci --color=yes
+common:ci --config=gcc
```
jmr: We should select gcc in the ci config.  macOS will use bazel.

CMakeLists.txt
```diff
@@ -119,23 +119,30 @@ if (MSVC)
...
-    # Avoid megabytes of warnings like:
-    # util/math/vector.h:178:16: warning: optimization attribute on
-    # ‘double sqrt(double)’ follows definition but the attribute doesn’t
-    # match [-Wattributes]
-    add_definitions(-Wno-attributes)
-    add_definitions(-Wno-deprecated-declarations)
-    add_definitions(-Wno-nullability-completeness)
-    # Suppress noisy AArch64 ABI notes about parameter passing changes in
-    # GCC 10.1.  Irrelevant since we don't mix objects from old compilers.
-    # TODO: Use check_cxx_compiler_flag(-Wno-psabi) to guard this.
```
jmr, line 131: Actually, this was a silly TODO.  it's simpler just to depend on -Wno- doing nothing.

jmr (changes requested):

src/.bazelrc
abhinavgautam01: Added --cxxopt=-Wno-deprecated-declarations and --cxxopt=-Wno-sign-compare to build:gcc (same pair on build:clang for macOS).

CMakeLists.txt
abhinavgautam01: Restored the original multi-line comments for -Wattributes, nullability, psabi and -fsized-deallocation.

CMakeLists.txt
abhinavgautam01: Dropped the OSS #458 / “silence churn” intro and all check_cxx_compiler_flag uses (and the unused include(CheckCXXCompilerFlag)).

src/.bazelrc
abhinavgautam01: Replaced common:ci --config=gcc with build:ci --config=gcc so the CI config selects GCC for builds; comment notes Linux CI vs --config=clang on macOS.

CMakeLists.txt
abhinavgautam01: Removed the TODO about check_cxx_compiler_flag(-Wno-psabi); -Wno-psabi is always passed (unsupported toolchains can ignore it).

abhinavgautam01:
ping @jmr

PR #625, opened 2026-05-12
pybind: pos validation and class attribute constants for S2CellId bindings

deustis:
## Summary

- Add `S2CellId::kMaxPosition` constant (`2^61 - 1`) to `s2cell_id.h` with a C++ unit test verifying valid and overflow behavior
- Validate `pos` argument in `S2CellId.from_face_pos_level` Python binding; expose `MAX_POSITION` as a Python class attribute
- Switch Python binding constants to `cls.attr()` pattern for uniform style regardless of whether constants come from `static constexpr` members or enum values; document in README
- Fix BUILD: add missing `@abseil-cpp//absl/strings` dep to `s2cell_id_bindings`

### DIFF
--- a/src/s2/s2cell_id.h
+++ b/src/s2/s2cell_id.h
@@ -99,7 +99,8 @@ class S2CellId {
   static constexpr int kMaxLevel =
       S2::kMaxCellLevel;  // Valid levels: 0..kMaxLevel
   static constexpr int kPosBits = 2 * kMaxLevel + 1;
-  static constexpr uint64_t kMaxPosition = (~uint64_t{0}) >> kFaceBits;
+  static constexpr uint64_t kMaxPosition =
+      (~static_cast<uint64_t>(0)) >> kFaceBits;
   static constexpr int kMaxSize = 1 << kMaxLevel;
 
   explicit constexpr S2CellId(uint64_t id) : id_(id) {}
--- a/src/s2/s2cell_id.h
+++ b/src/s2/s2cell_id.h
@@ -99,6 +99,7 @@ class S2CellId {
   static constexpr int kMaxLevel =
       S2::kMaxCellLevel;  // Valid levels: 0..kMaxLevel
   static constexpr int kPosBits = 2 * kMaxLevel + 1;
+  static constexpr uint64_t kMaxPosition = (~uint64_t{0}) >> kFaceBits;
   static constexpr int kMaxSize = 1 << kMaxLevel;
 
   explicit constexpr S2CellId(uint64_t id) : id_(id) {}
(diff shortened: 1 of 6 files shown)


deustis:
@jmr, couple of fast follow improvements here to be able to validate the position attribute,  and I also found a way to use python class attributes for constants as you had suggested previously.

src/s2/s2cell_id.h
```diff
@@ -99,6 +99,7 @@ class S2CellId {
   static constexpr int kMaxLevel =
       S2::kMaxCellLevel;  // Valid levels: 0..kMaxLevel
   static constexpr int kPosBits = 2 * kMaxLevel + 1;
+  static constexpr uint64_t kMaxPosition = (~uint64_t{0}) >> kFaceBits;
```
jmr: this line is confusing swig. use `static_cast` or `#ifndef SWIG`

src/s2/s2cell_id_test.cc
```diff
@@ -101,6 +101,29 @@ TEST(S2CellId, FromFace) {
...
+
+  // pos values above kMaxPosition overflow the position field into the face
+  // bits, producing an invalid cell.
+  //
+  // Replicate FromFacePosLevel to avoid its is_valid() DCHECK:
+  //   S2CellId cell((static_cast<uint64_t>(face) << kPosBits) + (pos | 1));
+  constexpr int face = 5;
+  constexpr uint64_t kOverflowPos = S2CellId::kMaxPosition + 1;
+  S2CellId overflow_id(
+      (static_cast<uint64_t>(face) << S2CellId::kPosBits) + (kOverflowPos | 1));
```
jmr: here you can use `uint64_t{face}`

src/s2/s2cell_id_test.cc
```diff
@@ -101,6 +101,29 @@ TEST(S2CellId, FromFace) {
   }
 }
 
+TEST(S2CellId, MaxPosition) {
```
jmr: Split into two test cases: max position is valid, larger than max position is invalid

src/s2/s2cell_id.h
deustis: Done, switched to `static_cast<uint64_t>`.

src/s2/s2cell_id_test.cc
deustis: Done.

src/s2/s2cell_id_test.cc
deustis: Done, split into `MaxPositionIsValid` and `PositionAboveMaxIsInvalid`.

deustis:
@jmr, ready for another look here (also #626 and #627)

PR #626, opened 2026-05-13
pybind: Add S1ChordAngle bindings

deustis:
Add pybind11 bindings for S1ChordAngle.

Includes constructors, factory methods (from_radians, from_degrees, zero, right, straight, infinity, negative), properties (radians, degrees, e5/e6/e7), predicates (is_zero, is_negative, is_infinity, is_special, is_valid), geometric operations (to_angle, sin, cos, tan), arithmetic operators (+, -, +=, -=), comparisons, hash, and __repr__/__str__.

The `e5`/`e6`/`e7` properties, `sin`/`cos`/`tan`, and arithmetic operators (`+`, `-`, `+=`, `-=`) raise `ValueError` for special values (`negative()` or `infinity()`). Operations such as comparisons, `radians`, `degrees`, `to_angle`, and string representations are intentionally unrestricted.

(Part of a series addressing https://github.com/google/s2geometry/pull/524)

### DIFF
--- a/src/python/s1chord_angle_bindings.cc
+++ b/src/python/s1chord_angle_bindings.cc
@@ -2,25 +2,23 @@
 #include <pybind11/operators.h>
 
 #include <cmath>
-#include <cstdint>
 #include <sstream>
 
 #include "absl/strings/str_cat.h"
 #include "s2/s1angle.h"
 #include "s2/s1chord_angle.h"
 #include "s2/s2point.h"
+#include "s2/s2pointutil.h"
 
 namespace py = pybind11;
 
 namespace {
 
 void MaybeThrowNotUnitLength(const S2Point& p, const char* name) {
-  // Use a generous tolerance; this matches the spirit of the C++
-  // ABSL_DCHECK(S2::IsUnitLength(x)) which uses ~4 * DBL_EPSILON.
-  double norm = p.Norm();
-  if (std::abs(norm - 1.0) > 1e-12) {
+  if (!S2::IsUnitLength(p)) {
     throw py::value_error(
-        absl::StrCat(name, " must be a unit-length vector (norm=", norm, ")"));
+        absl::StrCat(name, " must be a unit-length vector (norm=",
+                     p.Norm(), ")"));
   }
 }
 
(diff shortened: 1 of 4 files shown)


jmr:
> Special values (negative(), infinity()) are guarded in all operations where they are not meaningful.

Such as which? And what is "guarded"? Throws an exception?

deustis:
Specifically, `e5`/`e6`/`e7`, `sin`/`cos`/`tan`, and the arithmetic operators (`+`, `-`, `+=`, `-=`) call `MaybeThrowIfSpecial` and raise `ValueError` if either operand is `negative()` or `infinity()`. Operations like comparisons, `radians`, `degrees`, `to_angle`, and string methods are intentionally unrestricted. I've updated the description to be more precise.

src/python/s1chord_angle_bindings.cc
```diff
@@ -0,0 +1,202 @@
...
+           "Add two chord angles, clamping the result to [0, Pi].\n\n"
+           "Raises ValueError if either operand is Negative() or Infinity().")
+      .def("__sub__", [](const S1ChordAngle& a, const S1ChordAngle& b) {
+               MaybeThrowIfSpecial(a, "left operand");
+               MaybeThrowIfSpecial(b, "right operand");
+               return a - b;
+           }, py::is_operator(),
+           "Subtract two chord angles, clamping the result to [0, Pi].\n\n"
+           "Raises ValueError if either operand is Negative() or Infinity().")
+      .def("__iadd__", [](S1ChordAngle& self, const S1ChordAngle& other) {
```
jmr: I think these should be removed, since they can mutate something in a dict.  is that right? there may be other classes with this problem.

src/python/s1chord_angle_test.py
```diff
@@ -0,0 +1,276 @@
...
+        b = s2.S1ChordAngle.from_degrees(60.0)
+        diff = a - b
+        self.assertTrue(diff.is_zero())
+
+    def test_sub_special_raises(self):
+        a = s2.S1ChordAngle.from_degrees(30.0)
+        with self.assertRaises(ValueError):
+            _ = a - s2.S1ChordAngle.negative()
+
+    def test_iadd(self):
```
jmr: this should still work, so give it a different name

src/python/s1chord_angle_bindings.cc
deustis: Done.

Removed `__iadd__` and `__isub__`. Without them, `+=`/`-=` fall back to `__add__`/`__sub__` + rebinding, so the operations still work but no longer mutate the object in place. Renamed the corresponding tests to `test_add_assign`/`test_sub_assign` accordingly.

The same issue applies to the other hashable types in PR #627 (`S1Angle`, `R2Point`, `S2LatLng`, `S2Point`). I have updated that PR to remove some of the in-place operators and clarify in the README

src/python/s1chord_angle_test.py
deustis: Renamed to `test_add_assign` and `test_sub_assign`.

PR #681, opened 2026-09-18
Validate loop_starts when decoding a lax polygon shape

sushant-me:
Split out of #675, per the request there. Independent of #680.

### The bug

`chain_edge()` computes `vertices_[loop_starts_[i] + j]`, and neither
`S2LaxPolygonShape::Init` nor `EncodedS2LaxPolygonShape::Init` validated the
offsets they had just decoded. An encoding whose `loop_starts` are inconsistent
with the vertex array therefore reads outside it.

Confirmed with AddressSanitizer rather than inferred — the read is real but does
not fault in a normal build:

```
ERROR: AddressSanitizer: heap-buffer-overflow
READ of size 24
  #0 EncodedS2PointVector::At               src/s2/encoded_s2point_vector.h:166
  #1 EncodedS2PointVector::operator[]       src/s2/encoded_s2point_vector.h:183
  #2 EncodedS2LaxPolygonShape::chain_edge   src/s2/s2lax_polygon_shape.h:306
  #3 EncodedS2LaxPolygonShape::edge         src/s2/s2lax_polygon_shape.cc:360
```

### The change

Require the offsets to start at 0, be non-decreasing, and end at the vertex
count, in both the eager and the lazy decode path.

### Verification

The test builds a valid two-loop shape, encodes it, and then flips only the
"starts at 0" rule — leaving the offsets non-decreasing and ending at the vertex
count, so a check covering only monotonicity or only the final offset still
passes it. It asserts the layout of the tail before corrupting it, so a change
to the encoding fails loudly instead of silently patching the wrong byte.

```
$ ./s2lax_polygon_shape_test --gtest_filter='*RejectsInconsistentLoopStarts*'

before:  [  FAILED  ] 1 test.
after:   [  PASSED  ] 1 test.
```

Asserting on `Init()`'s return value rather than on a traversal is deliberate:
an out-of-bounds read does not fault without a sanitizer, so a test that only
walked the shape would pass both before and after.

Suites on this branch: `s2lax_polygon_shape_test` 21/21,
`encoded_s2shape_index_test` 9/9, `mutable_s2shape_index_test` 29/29.


sushant-me:
Note on the cancelled check: `cmake (macos-15-intel)` ended with

```
##[error]The operation was canceled.
```

inside the `cmake --build . --parallel 4` step — the runner was cancelled mid-build (the job ran 17m6s).
The test steps in that same job then reported `100% tests passed out of 1878` and `100% tests passed out
of 20`, and `lint` and both workflow-security jobs are green. So the red X is the runner, not the change.

I pushed an empty commit to re-trigger the workflow (no source change); the new runs are waiting on
approval. I cannot re-run jobs myself on this repository, so if it is more convenient to just hit
"Re-run failed jobs" on the old run, that would be equivalent.

